PR #13, opened 2022-09-28
refactor: `Slider` now manages its value via state

trickypr:
Useful for syncing state between restarts, like [here](https://github.com/trickypr/carpet/blob/0e7868d8fc7c8b9313a460c91ce53eff178bf87c/src/frontend/sound.rs#L34).

### DIFF
--- a/components/src/slider.rs
+++ b/components/src/slider.rs
@@ -3,30 +3,30 @@ use freya_elements as dioxus_elements;
 
 #[derive(Props)]
 pub struct SliderProps<'a> {
-    pub onmoved: EventHandler<'a, f64>,
-    pub width: f64,
-    /// The value that the slider will use when first initialized.
     #[props(optional)]
-    pub starting_value: Option<f64>,
+    pub onmoved: Option<EventHandler<'a, f64>>,
+    pub width: f64,
+    pub value: &'a UseState<f64>,
 }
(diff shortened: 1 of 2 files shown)


trickypr:
I am going to change how this is structured, so I am going to mark this as a draft.

trickypr:
Cool, I have rewritten it so now the slider takes in a state and will update it on change. `onmoved` is still around (but optional) in case you need to update external data. If Dioxus has an equivalent of [`useEffect`](https://reactjs.org/docs/hooks-effect.html), it might be a good idea to replace `onmoved` with that, to ensure that state is always in sync.

marc2332:
I have made it fully controlled, I don't like allowing it to be controlled or uncontrolled, I think having two options would make it harder to use and confuse the users. 
 
I understand your argument about the common misuse of React forms. But this is not HTML, we do not inherit all the HTML elements, therefore we can make it fully controllable, which is usually the desired option. I am all ears if you have another idea, but I strongly believe the Slider should be controlled by default, because, what benefits do you really get from having an uncontrolled slider?

trickypr:
Seems fair, I am fine with it for the moment

marc2332:
Okay, I am merging this. As I said, I am open to new ideas <3

PR #22, opened 2022-10-07
feat: Labels will now fit it's size according to it's text

marc2332:
Labels will now fit it's size according to its text, I accomplished this by using Skia's Paragraph. WIP

Example:

image

image


marc2332:
image

image

PR #103, opened 2023-03-09
Update macOS setup description

wusyong:
It seems macOS doesn't require any setup. Maybe this can reflect to the doc.


marc2332:
Awesome :magic_wand:

Issue #1, opened 2022-08-30
bug: High DPI not supported


trickypr:
Application text is too small when compared to other system applications.

<img width="777" alt="Screen Shot 2022-08-30 at 8 31 45 pm" src="https://user-images.githubusercontent.com/23250792/187414950-ad34a46c-1d5f-4fa6-9bed-70b7cebe5aa7.png">

Left: Example application
Right: Color meter for text size reference.

marc2332:
I think this is easy fix.

marc2332:
hi @trickypr do you mind trying https://github.com/marc2332/freya/pull/146 ? This time it should work!

PR #146, opened 2023-04-07
feat: Adaptive layout scale

marc2332:
Closes https://github.com/marc2332/freya/issues/1

It scales the layout with the window's scale factor


trickypr:
Screen Shot 2023-04-08 at 8 25 21 pm

Seems to be working correctly

PR #205, opened 2023-05-23
fix: Add copyright year in license

load1n9:


marc2332 (changes requested):
Also bring back the Cargo.lock file and remove it from `.gitignore`

PR #222, opened 2023-06-08
feat(breaking): Extended font styling

Tropix126:
This PR adds support for more methods of font styling. Currently, you have the option of specifying "bold", "italic", and "bold-italic" styles through the `font_style` property, which is the extent to which freya supports font styling. This PR introduces the `font_weight`, `font_slant`, and `font_width` properties.
- [x] `font_weight` allows weights of "invisible" through "extra-black" to be added to fonts, which replaces the previous `font_style` property's "bold/normal" mode.
- [x] `font_style` allows "normal", "italic", and "oblique" styles, replacing the previous `font_style` property's 	"italic" mode.
- [x] `font_width` controls the horizontal spacing between font characters.

This is a breaking change, since it removes the previous `font_style` setting. I can maybe find a way to keep backwards compatibility if needed.

image


marc2332:
Whoahhh! :open_mouth:  Amazing work @Tropix126 :clap: 

Regarding the deprecation of `font_style`, what if we actually kept it, but made it behave as your new `font_slant` ? It would be exactly the same as CSS's [`font_style`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-style)

state/src/font_style.rs
```diff
@@ -31,6 +36,12 @@ impl FontStyle {
             ..FontStyle::default()
         }
     }
+
+    pub fn to_skia_font_style(&self) -> skia_safe::font_style::FontStyle {
```
marc2332, line 40: Maybe this could be made by using the `From` trait?

```rust
impl From<FontStyle> for skia_safe::font_style::FontStyle {
    fn from(value: FontStyle) -> Self {
        skia_safe::font_style::FontStyle::new(value.font_weight, value.font_width, value.font_slant)
    }
}
```

marc2332 (changes requested):

state/src/font_style.rs
Tropix126: Should be done, let me know if anything I changed is wrong.

state/src/font_style.rs
```diff
@@ -162,12 +191,63 @@ impl State<CustomAttributeValues> for FontStyle {
...
+        "100" => Weight::THIN,
+        "200" => Weight::EXTRA_LIGHT,
+        "300" => Weight::LIGHT,
+        "400" => Weight::NORMAL,
+        "500" => Weight::MEDIUM,
+        "600" => Weight::SEMI_BOLD,
+        "700" => Weight::BOLD,
+        "800" => Weight::EXTRA_BOLD,
+        "900" => Weight::BLACK,
+        // "950" => Weight::EXTRA_BLACK,
```
marc2332, line 231: Shouldn't we enable both this one and the `// "50" => skia_safe::font_style::Weight::INVISIBLE,` ?

Tropix126: Probably. I left it commented because its not really something in the OpenType standard, but CSS does use weights of 950 (not 50 though skia does let you). I'll uncomment though since it makes sense and I wanted some outside feedback.

marc2332:
Amazing work @Tropix126 👏 Thank you agan 😄 🦀

PR #230, opened 2023-06-11
feat: Initial implementation of borders

Tropix126:
This PR adds a new `border` property that lets you add outlines to `rect`/`container` elements. It's an initial implementation, so its missing some aspects that can be added in later, notably:
- Physical pixel alignment for hairline borders (borders might not render as clearly on devices with fractional scaling due to antialiasing and subpixel boundaries, see https://github.com/flutter/flutter/issues/116278#issue-1469661628).
- Separate border colors/widths for one side. `border_{top,left,bottom,right}` can be added later, but that goes a little over my head right now.
- Border styles other than `solid` or `none`. `dashed` and `dotted` styles can be implemented using `PathEffect::dash`. The other ones seen in CSS require some more advanced rendering.
- Borders affecting layout. In CSS, borders are part of the box model, so they effectively act as an addition to `padding` when present (and `box-sizing` is `border-box`). This shouldn't be a huge issue, as padding can be manually added for now if they're needed. It might be better to keep it like this anyways, since design tools like figma don't do this either.

It also refactors some of the `rect/container` rendering code which will make it easier to add per-side corner-radius support in the future possibly.

image


marc2332:
Amazing work @Tropix126 Thank you for your contribution ❤️

PR #236, opened 2023-06-13
feat: CSS-like shadows

Tropix126:
This PR extends `shadow` to support most features in the [CSS Backgrounds and Borders Module Level 3](https://www.w3.org/TR/css-backgrounds-3/#propdef-box-shadow) `box-shadow` specification.
- [x] Correct(?) rendering of blur radius, x/y offset, and spread radius.
- [x] Inset Shadows
- [x] Multiple Shadows
- [x] Parsing (this is rather sketchy currently and should probably be refactored at some point)

image

Also i've temporarily removed `shadow` from DevTools since it'd break things otherwise.


marc2332 (approved):
🚀

marc2332:
Thank you again @Tropix126 😄

PR #241, opened 2023-06-18
feat: Text shadows

Tropix126:
This PR adds support for basic text shadows. Skia has its own `TextShadow` API that isn't as powerful as the custom drawing we use on `rect`/`container`, but unfortunately that drawing isn't possible to perform on paragraphs currently. This means:
- No support for spread radius (shadow intensity).
- No support for inset shadows.
- Text stroking (borders) likely isn't viable unless skia updates their APIs or I find a way to convert a `Paragraph` to a `TextBlob`.

image

This also adjusts the `parse_borders` function to be more general (`split_borders`), and moves the actual parsing into respective state structs to accomodate for two different shadow types.


elements/src/elements.rs
```diff
@@ -203,6 +203,7 @@ builder_constructors! {
     };
     label {
         color: String,
+        shadow: String,
```
marc2332, line 206: Let's name it `text_shadow`

state/src/style.rs
```diff
@@ -90,7 +90,14 @@ impl State<CustomAttributeValues> for Style {
                     }
                     "shadow" => {
                         if let Some(value) = attr.value.as_text() {
-                            style.shadows = parse_shadows(value, *scale_factor)
+                            style.shadows = split_shadows(value)
```
marc2332, line 93: Why was `parse_shadows` removed?

marc2332 (changes requested):

state/src/style.rs
Tropix126: I renamed it to `split_shadows` and split off the parsing logic so that the splitting could be shared in between `TextShadow` and `Shadow`, which each have their own parsers.

marc2332 (approved):

marc2332:
before merging I will push some documentation changes to the book so the new attribute is documented too

marc2332:
> before merging I will push some documentation changes to the book so the new attribute is documented too

I also need to add it to the devtools

marc2332 (approved):

marc2332:
Nice work @Tropix126  👏  👏  👏

PR #245, opened 2023-06-19
feat: Pixel snapping

Tropix126:
Implements snapping for `rect` and `container` elements to pixel boundaries, allowing for sharper rendering of borders and rect edges.
- This is an initial pass implementation. In the future, this might be better suited to go in torin.


crates/renderer/src/renderer.rs
```diff
@@ -83,21 +83,23 @@ pub fn render_skia(
             }
         }
 
+        let snapped_area = area.round();
```
marc2332, line 86: Does it really make sense to round these areas? For example, browsers do not round up paddings

marc2332 (changes requested):

crates/renderer/src/renderer.rs
Tropix126: It's definetly a sensible choice for `rect`, although it might not be necessary for other element types:

(without `snapped_area`)
image
image

(with `snapped_area`)
image
image

This seems to also have a *very slightly* improved effect on image rendering as well, but only on the edges in some rare cases. For the other elements, it's probably not a bad idea to round either way for edge cases like text decorations and SVG rendering.

marc2332: I'll see what I can do because I am getting some weird paddings on some components like the Switch

Tropix126: Yeah,  vertical centering in places like switches can cause issues. This is a problem in browsers too iirc, and requires either having blurry edges for the switch handle or even spacing between the switch height and switch handle height.

marc2332: > Yeah, vertical centering in places like switches can cause issues. This is a problem in browsers too iirc, and requires either having blurry edges for the switch handle or even spacing between the switch height and switch handle height.

2.6px
image

3px
image


comparison 2.6px vs  3px

image

Mmm, do browsers actually round the pixels? This doesn't feel like they do, or at least, the same way you do

Tropix126:
Closing this in favor of a new PR.

PR #256, opened 2023-06-24
feat: Linear gradient fill

Tropix126:
This is an initial implementation of a linear gradient fill mode.
image

They can currently be used to fill `border`, `shadow`, and `background`, but I hope to find a way to also style text properties.
image

Differences between CSS gradients:
- Color stop positions are required. This might change at a later time.
	- e.g. `linear-gradient(red, blue)` is not valid. Needs to be `linear-gradient(red 0%, blue 100%)`
- Color stop positions can only be specified using percentages of the gradient line, not using length units.
- Gradient rotation is counterclockwise, not clockwise.

This is a pretty big change under the hood. Devtools are broken because of changes to node state, and tests need rewriting still. Some parsing had to be changed for other properties too (mainly due to the new splitter functions).


marc2332:
We can write the docs in another PR

marc2332 (approved):

PR #265, opened 2023-06-29
refactor(breaking): Consistency for rotation units

Tropix126:
This is an initial fix-up of some inconsistencies in freya's styling values.
- Removes the "unitless" form of `hsl` colors (`hsl(45, 0.5, 0.1, 0.2, 0.3)`), instead preferring the more explicit `hsl(45deg, 50%, 10%, 20%, 30%)` syntax.
- Makes `rotate` attribute use `deg` units.

This makes it consistent with the required units in `linear-gradient`. Let me know if any of this needs further discussion or whatnot.


marc2332:
Nice! Thank you :D

PR #274, opened 2023-07-04
feat: `text_overflow`

Tropix126:
Adds a new `text_overflow` attribute for `paragraph` and `label` elements, which allows people to adjust the behavior of text (either to clip it or show ellipsis) when it exceeds its `max_lines`. The default is `"ellipsis"` which cuts off the text and appends a `…` character to the end.

This also fixes support for `max_lines` on `label` elements (since I ran into that when implementing this).

https://github.com/marc2332/freya/assets/42101043/c05b4227-10b0-432b-a980-11c728ed6517


book/src/guides/font_style.md
```diff
@@ -2,21 +2,22 @@
...
-- [`font_width`](#font_width)
-- [`line_height`](#line_height)
-- [`max_lines`](#max_lines)
-- [`letter_spacing`](#letter_spacing)
-- [`word_spacing`](#word_spacing)
-- [`decoration`](#decoration)
-- [`decoration_style`](#decoration_style)
-- [`decoration_color`](#decoration_color)
-- [`text_shadow`](#text_shadow)
+- [Font Style](#font-style)
```
marc2332, line 5: Let's keep it as it was before

marc2332 (changes requested):

marc2332 (approved):

PR #299, opened 2023-09-09
Virtual scroll rework

kobutri:
# rework the virtual scroll view

## Main features
- allow for dynamic and multiple item sizes
- allow scrolling elements partially instead of snapping the closed element to the top
- give scroll thumbs a minimal size

## Known Problems
- when the window is resized it sometimes freezes up

### DIFF
--- a/crates/components/src/scroll_views/mod.rs
+++ b/crates/components/src/scroll_views/mod.rs
@@ -55,10 +55,10 @@ pub fn get_scrollbar_pos_and_size(
         inner_size
     } else {
         let viewable_ratio_height = viewport_size / inner_size;
-        viewport_size * viewable_ratio_height
+        (viewport_size * viewable_ratio_height).max(30.0)
     };
-    let scroll_position = (100.0 / inner_size) * -scroll_position;
-    let scrollbar_position = (scroll_position / 100.0) * viewport_size;
+    let scroll_position = (100.0 / (inner_size-viewport_size)) * -scroll_position;
+    let scrollbar_position = (scroll_position / 100.0) * (viewport_size - scrollbar_height);
     (scrollbar_position, scrollbar_height)
 }
 
--- a/crates/components/src/scroll_views/virtual_scroll_view.rs
+++ b/crates/components/src/scroll_views/virtual_scroll_view.rs
@@ -22,10 +22,9 @@ type BuilderFunction<'a, T> = dyn Fn(
     ),
 ) -> LazyNodes<'a, 'a>;
 
-#[derive(Clone)]
 pub enum ItemSize {
     Fixed(f32),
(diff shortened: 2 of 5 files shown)


marc2332:
I must admit I am quite confused by these `SizeCache`, `Node`, `InnerNode`, etc, do you mind writing an overview of the logic? Seems like too complex imo

kobutri:
@marc2332 
You are right. It's a rather confusing and ad hoc implementation. I'll rework it once the idea has been validated.
So whats going on?

## ItemSize
We want to allow users specify the item height (or width for horizontal) in multiple ways .This is done via the ItemSize enum. 
It contains FixedSize for when all items have the same size, PerItem when every item has a potentially different size, but it is known ahead of time. Note that the function is only called for every index. This can't be used to specify size changes. The last one is Dynamic when the size isn't known ahead of time.

## SizeCache Motivation
In addition to that the VirtualListView should treat these item sizes as hints. And always measure the actual size. This is facilitated via the Wrapper component, that measures the actual size of the items and inserts them into a SizeCache.
So what does the SizeCache do? On the surface it has a very simple task, store the measured item sizes and given a scroll offset and viewport size produce a range of items that should be rendered a long with their vertical alignment inside the viewport.

## SizeCache structure
In order todo this efficiently the SizeCache employs a tree structure. The Nodes can either be inner nodes, leaf nodes or leaf range nodes. The first two should be self explanatory and the last one simply represents a range of nodes whose size has not been measured yet (for simplicity the PerItem sizes are treated as initial measurements). Each inner node stores the range of elements it represents, as well as their total measured size and the number of items covered by leaf range nodes (I'm looking for a better name :)). Each inner node contains at most chunk_size many elements. The the range lookup becomes just two tree searches, where for every level of the tree we only need to traverse their chunk_size many children and add up their measured size, as well as the estimated size of the elements covered by leaf range nodes.

## SizeCache implementation
Insertion becomes a pretty standard tree insertion. We first find their the leaf node to update or the leaf range node that contains the item. In the latter case the leaf range node is split into two and a leaf node inserted in the middle. From there it's simply a matter of updating the total measured size and leaf range count numbers in the parents and rebalance the tree.
All in all it's basically a B-Tree with some additional metadata attached to the inner nodes.

I hope this gives you a rough idea about whats going on.
It might very well be too complex, but I have yet to come up with an idea for a solution that is at least conceptually equally robust.

marc2332:
I am still thinking about this.

I like the following although I would need to check the exact implementation:
- Per item sizing
- Minimum size for thumbs
- "Smooth" scrolling

Don't like
- Dynamic sizing: too much code complexity and performance overhead (caching, `Wrapper`, etc), and little use cases

PR #301, opened 2023-09-13
fix: Manually request redraw after resize on macOS

kobutri:
When resizing really slowly it sometimes stops responding to user input.
The `resize` function on `App` clears the layout.
Normally the OS would automatically schedule a redraw, but at least on Mac OS this doesn't always happen.
Without calling another redraw the layout remains empty, making it impossible to process input events. 
Manually requesting a redraw from the `WindowEnv` `resize` function fixes the problem.

### DIFF
--- a/crates/renderer/src/window.rs
+++ b/crates/renderer/src/window.rs
@@ -271,6 +271,8 @@ impl<T: Clone> WindowEnv<T> {
             NonZeroU32::new(width.max(1)).unwrap(),
             NonZeroU32::new(height.max(1)).unwrap(),
         );
+
+        self.window.request_redraw();
     }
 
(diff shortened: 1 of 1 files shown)


marc2332:
Could this be a Winit bug? Or it's the expected behavior?

kobutri:
I did some more debugging and found, that while all resize events emit redraw events they weren't always in the expected order. The winit documentation is a little unclear about this, but from what I read on some winit issues, they are not providing any order guarantees for `RequestRedraw` events anymore.  The only thing we know for sure, is that calling `request_redraw` will trigger a redraw.

marc2332 (approved):

marc2332:
Many thanks! 🚀

PR #309, opened 2023-09-21
fix(docs): Update Torin's link in differences with Dioxus

tigerros:
The torin link was outdated. This updates it and also adds a permalink which will never "expire", unless the repository is deleted, or @marc2332 changes his name.


marc2332 (changes requested):
Nice find, but it shouldn't point to an old version, it should just point to https://github.com/marc2332/freya/tree/main/crates/torin

tigerros:
>  it should just point to https://github.com/marc2332/freya/tree/main/crates/torin

The files might get moved around or something and the link becomes invalid, that's why I thought a permalink would make sense. However, I noticed torin is already on crates.io, so I added a link to there instead of the permalink. Unless you change the name of the crate, it's not going to become invalid. Problem solved.

marc2332 (approved):

PR #318, opened 2023-09-25
feat: New torin benchmark

marc2332:
Depends on https://github.com/marc2332/freya/pull/314

New torin benchmark: deep + wide + cached


nicoburns:
@marc2332 These benchmark results looks suspiciously good to me. Specifically the `benchmarks/big trees (wide)` and `benchmarks/big trees (deep)` results where I'm assuming you're attempting to measure a fresh uncached layout? I think you might be making the same mistake that the initial version of Taffy's benchmarks did: reusing the same node tree across all iterations of the benchmarks, which means that if the benchmark framework does 100 iterations and averages the results, then only the first iteration computes layout and subsequent iteration just access the cached layout. So you're actually testing `(time_to_layout + (99 * time_to_access_cached_value)) / 100` which is probably ~= `time_to_layout / 100`

To fix this in Taffy, we had to switch from using `group.iter` to using `group.iter_batched` (see https://bheisler.github.io/criterion.rs/book/user_guide/timing_loops.html#iter_batchediter_batched_ref), which allowed us to construct a new tree for each iteration.

marc2332:
> @marc2332 These benchmark results looks suspiciously good to me. Specifically the `benchmarks/big trees (wide)` and `benchmarks/big trees (deep)` results where I'm assuming you're attempting to measure a fresh uncached layout? I think you might be making the same mistake that the initial version of Taffy's benchmarks did: reusing the same node tree across all iterations of the benchmarks, which means that if the benchmark framework does 100 iterations and averages the results, then only the first iteration computes layout and subsequent iteration just access the cached layout. So you're actually testing `(time_to_layout + (99 * time_to_access_cached_value)) / 100` which is probably ~= `time_to_layout / 100`
> 
> To fix this in Taffy, we had to switch from using `group.iter` to using `group.iter_batched` (see https://bheisler.github.io/criterion.rs/book/user_guide/timing_loops.html#iter_batchediter_batched_ref), which allowed us to construct a new tree for each iteration

Hey Nico!

I have two types of benchmarks, the ones where I create the layout on each iteration, so, nothing is cached. And the other one is where I create and cache the layout before the benchmark, and then I invalidate a certain element on each iteration. But... I agree that this benchmark from this PR in particular (and maybe the others) are too good, hence why I haven't merged it yet, so don't worry about it

I still have to improve these, it's my first time doing benchmarks! I'll take a look at `iter_batched` asap 😄

marc2332:
Looks like it is more accurate with iter_batched now @nicoburns ! Thanks for the suggestion! 😄

marc2332:
I still have to improve the benchmarks though! Some of them are not very realistic yet

nicoburns:
@marc2332 Benchmarking code looks reasonable to me now. Results are still very fast! (roughly 10x faster than Taffy (and Morphorm last time I saw numbers for that)). But Torin is quite a bit simpler than either (no stretch/flex functionality at all yet, right?), so that seems plausible to me.

Taffy's benchmarks evenly distribute the children in each layer rather than only giving one child in each layer children. But I don't think there's anything wrong with your approach. It's just different.

marc2332:
> (no stretch/flex functionality at all yet, right?)

Yeah, that's correct! Torin is way simpler

> Taffy's benchmarks evenly distribute the children in each layer rather than only giving one child in each layer children. But I don't think there's anything wrong with your approach. It's just different.

Exactly! I want to do it just like Taffy because I think it's more realistic, perhaps I could leave there but change the name so it becomes more obvious that there is only one child with children per layer

PR #325, opened 2023-10-08
feat: Password mode for `Input` component

OlshaMB:
### Adds
- InputIsHidden enum where state Hidden contains a char that is used to hide chars of the input
- If input is hidden we use rust repeat method on the obfuscation char
- Example using hidden input for password input(also includes ability to show and hide the password)
### Potential changes
- Change the enum to string*
### Potential bugs
- Wrong behavior with special chars like `\n` or `\t`**
-----------------------------------------------------------------------------------------------------
*This would make it not typesafe and introduce addtional checks to prevent bugs
**Should we block input of special chars or exclude some of the charecters from hiding

### DIFF
--- a/crates/components/src/input.rs
+++ b/crates/components/src/input.rs
@@ -10,11 +10,19 @@ use freya_hooks::{
 };
 use winit::window::CursorIcon;
 /// Enum to declare is [`Input`] hidden.
-pub enum InputIsHidden {
-    /// The input shown
+#[derive(Default)
+pub enum InputMode {
+    /// The input text is shown
+    #[default]
     Shown,
-    /// The input is obfuscated with a character
+    /// The input text is obfuscated with a character
     Hidden(char),
+}
+
+impl InputMode {
+    fn new_password() -> Self {
+        Self::Hidden('*')
+    }
 }
 /// [`Input`] component properties.
 #[derive(Props)]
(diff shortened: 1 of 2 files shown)


crates/components/src/input.rs
```diff
@@ -9,14 +9,23 @@ use freya_hooks::{
...
 };
 use winit::window::CursorIcon;
-
+/// Enum to declare is [`Input`] hidden.
+pub enum InputIsHidden {
+    /// The input shown
+    Shown,
+    /// The input is obfuscated with a character
+    Hidden(char),
+}
```
marc2332, line 18:
```suggestion
#[derive(Default)
pub enum InputMode {
    /// The input text is shown
    #[default]
    Shown,
    /// The input text is obfuscated with a character
    Hidden(char),
}

impl InputMode {
    fn new_password() -> Self {
        Self::Hidden('*')
    }
}
```

crates/components/src/input.rs
```diff
@@ -9,14 +9,23 @@ use freya_hooks::{
...
 /// [`Input`] component properties.
 #[derive(Props)]
 pub struct InputProps<'a> {
     /// Current value of the Input
     pub value: String,
     /// Handler for the `onchange` event.
     pub onchange: EventHandler<'a, String>,
+    /// Is input hidden with a character. By defaultv input is shown
+    #[props(default = InputIsHidden::Shown, into)]
+    hidden: InputIsHidden,
```
marc2332, line 28:
```suggestion
    /// Is input hidden with a character. By default input text is shown.
    #[props(default = InputMode::Shown, into)]
    hidden: InputMode,
```

marc2332 (changes requested):
Nice!! Just some code clean up

marc2332:
Many thanks!! ✨

PR #330, opened 2023-10-14
feat: Enhanced layout alignments

marc2332:
Info https://github.com/marc2332/freya/issues/327

# Changes
- [x] New `main_align` and `cross_align` attributes
   - Values: `start` (default), `center` and `end`
- [x] Deprecate the `both` direction 
- [x] Deprecate the `display` attribute
- [x] Update book
- [x] Update examples
- [x] Refactored and cleaned up a lot of code
- [x] Updated and added new tests

Demo of `alignment.rs`:

image

PR #332, opened 2023-10-15
chore: Rename 'align' attribute to 'text_align'

marc2332:
To avoid confusion with the new alignment attributes from https://github.com/marc2332/freya/pull/330 , I think it would be better to rename the text alignment `align` attribute to `text_align`


marc2332:
Thoughts on this @OlshaMB ?

OlshaMB:
> Thoughts on this @OlshaMB ?

The idea of renaming to text_align is great, but also it breaks existing code

marc2332:
> > Thoughts on this @OlshaMB ?
> 
> The idea of renaming to text_align is great, but also it breaks existing code

We are already breaking code by removing the `display` property and removing the `both` direction, hehe

OlshaMB:
- In spec I deprecated it not removed, deprecation means that devs should migrate to new properties util next major release where it can be removed, I don't now what do we mean here by a major version, because the project hasn't reached 1.0.0, I think if you release the layout changes then we release in the next major which in case of beta is 0.x, so removing and breaking changes should be fine, but I think we should set the policy of removing only in major versions and deprecating in minor(maybe we should write contributing.md)

OlshaMB:
> In spec I deprecated, it not removed, deprecation means that devs should migrate to new properties util next major release where it can be removed, I don't now what do we mean here by a major version, because the project hasn't reached 1.0.0, I think if you release the layout changes then we release in the next major which in case of beta is 0.x, so removing and breaking changes should be fine, but I think we should set the policy of removing only in major versions and deprecating in minor(maybe we should write contributing.md)
**in short:**
Renaming and removing should be fine if we do it in next major.

marc2332:
> In spec I deprecated, it not removed, deprecation means that devs should migrate to new properties util next major release where it can be removed, I don't now what do we mean here by a major version, because the project hasn't reached 1.0.0, I think if you release the layout changes then we release in the next major which in case of beta is 0.x, so removing and breaking changes should be fine, but I think we should set the policy of removing only in major versions and deprecating in minor(maybe we should write contributing.md)

I plan to release this in a 0.2 version. Maintaining support with the new alignments and still support the old `both` direction and `display` would kinda hard in terms of maintaining the code tbh

OlshaMB:
As I said in my comment the versioning is still pre 1.0, so it's fine to release this in 0.2, but not in 0.1.x

OlshaMB:
> > In spec I deprecated, it not removed, deprecation means that devs should migrate to new properties util next major release where it can be removed, I don't now what do we mean here by a major version, because the project hasn't reached 1.0.0, I think if you release the layout changes then we release in the next major which in case of beta is 0.x, so removing and breaking changes should be fine, but I think we should set the policy of removing only in major versions and deprecating in minor(maybe we should write contributing.md)
> 
> 
> 
> I plan to release this in a 0.2 version. Maintaining support with the new alignments and still support the old `both` direction and `display` would be kinda hard in terms of maintaining the code tbh 

I understand that it's difficult, I just said that breaking changes should be in majors

marc2332:
> As I said in my comment the versioning is still pre 1.0, so it's fine to release this in 0.2(as it's major), but not in 0.1.x(it's minor)

Yeah, I agree

PR #355, opened 2023-10-22
fix: Abandon context if not current when closing the app

marc2332:


crates/renderer/src/window.rs
```diff
@@ -46,6 +47,14 @@ pub struct WindowEnv<T: Clone> {
     pub(crate) window_config: WindowConfig<T>,
 }
 
+impl<T: Clone> Drop for WindowEnv<T> {
+    fn drop(&mut self) {
+        if !self.gl_context.is_current() {
```
tronical, line 52: I suggest to try to make the context current and only abandon if making it current fails.

Issue #371, opened 2023-11-05
enhancement: `Popup` / `Dialog` component


marc2332:
Blocked by https://github.com/marc2332/freya/pull/370

PR #372, opened 2023-11-06
feat: `Popup` component

marc2332:
Closes https://github.com/marc2332/freya/issues/371
Depends on https://github.com/marc2332/freya/pull/268
image


marc2332:
> How about renaming this to `Dialog`? That's what it's called in HTML and "popup" is IMO associated with the annoying kind.

Doesn't Dialog imply that there must be a dialog to be made? Like, accept, cancel, whatever? Popup seems more generic to me 🤔

marc2332:
> > > How about renaming this to `Dialog`? That's what it's called in HTML and "popup" is IMO associated with the annoying kind.
> > 
> > 
> > Doesn't Dialog imply that there must be a dialog to be made? Like, accept, cancel, whatever? Popup seems more generic to me 🤔
> 
> Sure, but I just thought that since HTML calls it that, it might be more familiar. Also, it might not always be a "popup", but perhaps a more permanent/complex "subwindow". Like it could have a whole form in it or something, but at that point I wouldn't call it a popup.

What about "Modal"?

marc2332:
I'll leave it as "Popup"

marc2332:
I should probably support closing popups with Esc

marc2332:
> I should probably support closing popups with Esc

I'll open an issue for this

PR #383, opened 2023-11-17
feat: `Sidebar` component

marc2332:
## To Do
- integrate theming
- UX and UI improvements


tigerros:
I'll take a look at this if that's alright

marc2332:
> I'll take a look at this if that's alright

Sure! Np, but it is still WIP

tigerros:
> > I'll take a look at this if that's alright
> 
> Sure! Np, but it is still WIP

I'll at least update it for the new theming system

tigerros:
I just realized that `SidebarItem` is basically a `Link`. I'll finish https://github.com/marc2332/freya/pull/411 and then I'll update the router + sidebar example.

marc2332:
> I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.

It isn't just a `Link` as `SidebarItem` has an optional onclick handler + theming built-in

tigerros:
> > I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.
> 
> It isn't just a `Link` as it has an optional onclick handler + style built-in

Right, but I would prefer to remove the `Link` functionality from it, since it's not always going to be used that way. I think using a dedicated `Link` component when needed makes more sense. Do you think you could merge https://github.com/marc2332/freya/pull/411?

marc2332:
> > > I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.
> > 
> > 
> > It isn't just a `Link` as it has an optional onclick handler + style built-in
> 
> Right, but I would prefer to remove the `Link` functionality from it, since it's not always going to be used that way. I think using a dedicated `Link` component when needed makes more sense. Do you think you could merge #411?

I'll think about it. I don't like having a Link component because you can't use it in combination with `Button` for example because there is no event bubbling. That's why I removed it and why #411 is marked as blocked

tigerros:
> > > > I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.
> > > 
> > > 
> > > It isn't just a `Link` as it has an optional onclick handler + style built-in
> > 
> > 
> > Right, but I would prefer to remove the `Link` functionality from it, since it's not always going to be used that way. I think using a dedicated `Link` component when needed makes more sense. Do you think you could merge #411?
> 
> I'll think about it. I don't like having a Link component because you can't use it in combination with `Button` for example because there is no event bubbling. That's why I removed it and why #411 is marked as blocked

`ExternalLink` exists though and I think it has the same issue as `Link`, doesn't it? Or what do you mean exactly?

marc2332:
> > > > > I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.
> > > > 
> > > > 
> > > > It isn't just a `Link` as it has an optional onclick handler + style built-in
> > > 
> > > 
> > > Right, but I would prefer to remove the `Link` functionality from it, since it's not always going to be used that way. I think using a dedicated `Link` component when needed makes more sense. Do you think you could merge #411?
> > 
> > 
> > I'll think about it. I don't like having a Link component because you can't use it in combination with `Button` for example because there is no event bubbling. That's why I removed it and why #411 is marked as blocked
> 
> `ExternalLink` exists though and I think it has the same issue as `Link`, doesn't it? Or what do you mean exactly?

Yes, same for `ExternalLink`

tigerros:
> > > > > > I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.
> > > > > 
> > > > > 
> > > > > It isn't just a `Link` as it has an optional onclick handler + style built-in
> > > > 
> > > > 
> > > > Right, but I would prefer to remove the `Link` functionality from it, since it's not always going to be used that way. I think using a dedicated `Link` component when needed makes more sense. Do you think you could merge #411?
> > > 
> > > 
> > > I'll think about it. I don't like having a Link component because you can't use it in combination with `Button` for example because there is no event bubbling. That's why I removed it and why #411 is marked as blocked
> > 
> > 
> > `ExternalLink` exists though and I think it has the same issue as `Link`, doesn't it? Or what do you mean exactly?
> 
> Yes, same for `ExternalLink`

Why can't `Link` be merged then?

marc2332:
> > > > > > > I just realized that `SidebarItem` is basically a `Link`. I'll finish #411 and then I'll update the router + sidebar example.
> > > > > > 
> > > > > > 
> > > > > > It isn't just a `Link` as it has an optional onclick handler + style built-in
> > > > > 
> > > > > 
> > > > > Right, but I would prefer to remove the `Link` functionality from it, since it's not always going to be used that way. I think using a dedicated `Link` component when needed makes more sense. Do you think you could merge #411?
> > > > 
> > > > 
> > > > I'll think about it. I don't like having a Link component because you can't use it in combination with `Button` for example because there is no event bubbling. That's why I removed it and why #411 is marked as blocked
> > > 
> > > 
> > > `ExternalLink` exists though and I think it has the same issue as `Link`, doesn't it? Or what do you mean exactly?
> > 
> > 
> > Yes, same for `ExternalLink`
> 
> Why can't `Link` be merged then?

Because similarly to externalLink, having these two without event bubbling is having a half-working solution

tigerros:
> Because similarly to externalLink, having these two without event bubbling is having a half-working solution

I know, but `ExternalLink` exists, so why not `Link`? They should either both exist, or neither of them should exist, because they have the same problem. Not including thing A for problem X makes no sense when there's thing B that also has problem X.

marc2332:
> > Because similarly to externalLink, having these two without event bubbling is having a half-working solution
> 
> I know, but `ExternalLink` exists, so why not `Link`? They should either both exist, or neither of them should exist, because they have the same problem. Not including thing A for problem X makes no sense when there's thing B that also has problem X.

Ooops, I didn't realize `ExternalLink` was already on main, my bad. I guess it doesn't hurt more to have `Link` as well then

marc2332:
Closing in favor of https://github.com/marc2332/freya/pull/513

PR #384, opened 2023-11-18
feat: Gallery example

marc2332:
Depends on https://github.com/marc2332/freya/pull/383


marc2332:
Maintaining a Gallery example would be burden, I prefer to have splitted examples

PR #388, opened 2023-11-19
feat: Clipboard integration in `use_editable`

marc2332:
Depends on https://github.com/marc2332/freya/pull/346
Closes https://github.com/marc2332/freya/issues/473


marc2332 (commented):
~I should probably use dioxus-std's clipboard integration and base this PR on https://github.com/marc2332/freya/pull/346~

marc2332 (commented):
I should use Command in macOS

PR #406, opened 2023-12-01
Fix wording in book

tigerros:
> so in this case, the `rect` width will be defined bu the width of `label`:

into:

> so in this case, the `rect` width will be equivalent to the width of `label`:


marc2332:
thank you 🔥

PR #408, opened 2023-12-02
Mention the fact that the book uses some APIs from the git version

tigerros:
I was going through the book and noticed that some APIs in it (e.g., `position`) are not in the crates.io version (currently `0.1.8`). This PR just adds a warning at the beginning of the book. I know there's a note saying that the book is not always accurate, but it's not very helpful and explicit in this particular case.


marc2332:
I would then remove my note and just leave yours, thoughts?

marc2332:
Oops, lol

tigerros:
Sure. I also added a more generic disclaimer to not fully trust the book, rather than just warn about "nightly" APIs being used.

PR #411, opened 2023-12-04
feat: Add `Link` component back

tigerros:
## Features

- Add `Link`, which works like Dioxus router's `Link`. Example code:

  ```rust
  rsx! {
      Link { to: Route::Foo {}, label { "Route default (no tooltip)" } }
      Link { to: Route::Bar {}, tooltip: AnchorTooltip::Custom("Bar".to_string()), label { "Custom tooltip" } }
      Link { to: "https://github.com/marc2332/freya", label { "URL default (URL value tooltip)" } }
      Link { to: "https://github.com/marc2332/freya", tooltip: AnchorTooltip::None, label { "URL no tooltip" } }
  }
  ```
- Optimize `ExternalLink` a bit.

## Notes

- This makes `ExternalLink` obsolete, except for theming.
- `Link` might be a better name.
- This doesn't add any styles or themes. I think components should add the minimum amount of style. They can be wrapped in a `rect` to add shadows, change color, etc., but it's impossible to remove styles. As for the themes, I just didn't bother adding them, which is why this is a draft.


marc2332:
I actually had a `Link` component some time ago, but the problem is that it didn't work well because Freya doesn't have  events bubbling [yet](https://github.com/marc2332/freya/pull/268), I am waiting for the Dioxus 0.5 release. Once that is done, we should definitely merge this! I would call the component `Link` instead of Anchor, yes.

tigerros:
New things:

- Renamed `Anchor` to `Link`.
- Added the event propagation warning.
- Removed `ExternalLink` and added theming from it to `Link`.
- Migrated all references of `ExternalLink` to `Link`.

tigerros:
New changes:
- Added `ExternalLink` back for situations without `dioxus-router`.
- Made the example work.
- Updated, fixed, polished.

marc2332:
I'll get this working with the changes from https://github.com/marc2332/freya/pull/268

marc2332:
Amazing work @tigerros ! I just resolved all the conflicts and cleaned some code, this LGTM. I would deprecate `ExternalLink` in this same PR, thoughts?

marc2332:
> > Amazing work @tigerros ! I just resolved all the conflicts and cleaned some code, this LGTM. I would deprecate `ExternalLink` in this same PR, thoughts?
> 
> I don't think `ExternalLink` should be deprecated because `Link` only works with Dioxus's router and they do the same thing with external routes, with `Link` having some overhead.

I can try to make `Link` work when dioxus-router is not being used, it shouldn't be that hard

marc2332:
> > > > Amazing work @tigerros ! I just resolved all the conflicts and cleaned some code, this LGTM. I would deprecate `ExternalLink` in this same PR, thoughts?
> > > 
> > > 
> > > I don't think `ExternalLink` should be deprecated because `Link` only works with Dioxus's router and they do the same thing with external routes, with `Link` having some overhead.
> > 
> > 
> > I can try to make `Link` work when dioxus-router is not being used, it shouldn't be that hard
> 
> Did something change with hooks? I thought you had to call them every time so you can't remove `use_navigator`.

Nothing changed, but I can get the navigator myself in L138 instead of using the hook:

```rs
let router = try_consume_context::<RouterContext>()
            .expect("Must be called in a descendant of a Router component");

router.push(to.clone());
```

marc2332:
Just found a smalllllllllll bug with the events system in Freya, will open a PR later, after that, I will push some final touches and this PR will be ready to go

marc2332:
works!

marc2332:
I'll add a unit tests before merging

marc2332:
Nice work @tigerros !

Issue #423, opened 2023-12-07
enhancement: More flexible built-in component styling


tigerros:
I noticed that the built-in components don't have very good style configuration. There's PRs like [#422](https://github.com/marc2332/freya/pull/422) that add more style properties to components, but I  think that's not sustainable. Could there be a better way than just adding more style properties? Maybe a stylesheet?

marc2332:
That's what Themes are for, they are kind of a stylesheet. Although they mostly contain visual styling and not layout styling, perhaps I could also move the layout styling (margin, padding, etc) to Themes? What do you think?

tigerros:
> That's what Themes are for, they are kind of stylesheets.

Using themes is annoying when you try to change the style of just one component instead of the entire app or all of the `render!` macro contents. I think components could have an optional `theme` property that would override the inherited theme. For example:

```rust
render! {
    Button {
        theme: ButtonTheme {
            background: "red",
            ..LIGHT_THEME.button
        }
    }
}
```

> I could also add the layout styling (margin, padding, etc) to Themes (beside passing via prop)? What do you think?

Yeah I think layout should be in the themes. The split is pretty arbitrary, and it doesn't seem to be followed properly :P (#422 added `shadow` and `corner_radius` properties which are not layout) And if you go forward with my `theme` property suggestion, it would also mean that all style properties would be separated from the "functional" properties.

I also want to point out that the components don't fully utilize themes. As in, there's some properties which just have a hardcoded value (e.g., `border` is always `1 solid` in `Input`). That's not really an API issue though.

marc2332:
> ```rust
> render! {
>     Button {
>         theme: ButtonTheme {
>             background: "red",
>             ..LIGHT_THEME.button
>         }
>     }
> }
> ```

Yeah, I think I like that, I'll see what I can do



> Yeah I think layout should be in the themes. The split is pretty arbitrary, and it doesn't seem to be followed properly :P (#422 added `shadow` and `corner_radius` properties which are not layout) And if you go forward with my `theme` property suggestion, it would also mean that all style properties would be separated from the "functional" properties.

You are 💯 right

> I also want to point out that the components don't fully utilize themes. As in, there's some properties which just have a hardcoded value (e.g., `border` is always `1 solid` in `Input`). That's not really an API issue though.

I also noticed this yesterday when doing https://github.com/marc2332/freya/pull/422

tigerros:
> Yeah, I think I like that, I'll see what I can do

I would be happy to make a PR

PR #425, opened 2023-12-08
feat: Per-component themes

tigerros:
Closes https://github.com/marc2332/freya/issues/423

- Introduces a way to style individual components with individual styles.
- Made some properties public and cleaned some code.

Example:

```rust
render! {
    Button {
        theme: ButtonThemeWith {
            background: Some("blue".into()),
            font_theme: Some(FontThemeWith {
                color: Some("white".into()),
                ..Default::default()
            }),
            ..Default::default()
        },
        label { "Blue button" }
    }
}
```

Or with the `theme_with` macro:

```rust
render! {
    Button {
        theme: theme_with!(ButtonTheme {
            background: "blue".into(),
            font_theme: theme_with!(FontTheme {
                color: "white".into(),
            }),
        }),
        label { "Blue button" }
    }
}
```

This will only change the `background` and `color` of the button. Nothing else is affected. Subthemes are supported (e.g., `FontThemeWith`).

# Internals

Example of defining a theme:

```rust
// use_theme.rs
define_theme! {
    %[component]
    pub Test<'a> {
        %[cows]
        cow_string: str,
        %[borrowed]
        borrowed_data: &'a Foo,
        %[owned]
        owned_data: Bar,
        %[subthemes],
        font_theme: FontTheme,
    }
}
```

- The `%[component]` attribute will automatically generate docs that point to the component of the theme's name.

The following attributes are all field "separators". This means that you specify them only once per theme, and they will apply to all following fields until they encounter another separator. These must be in the same order, but you don't need to use all of them.

- `%[cows]` indicates that the next fields should be wrapped in `Cow<'static, _>`. This is more specific than the others, but it's an extremely common use case, because strings should always be wrapped in `Cow` to allow constant theme templates, but also dynamic theme overrides.
- `%[borrowed]` indicates that the next fields are borrowed data and should therefore be handled as such (this matters in the `apply_optional` function that the macro generates).
- `%[owned]` indicates that the next fields are owned data and should therefore be handled as such (this matters in the `apply_optional` function that the macro generates).
- `%[subthemes]` indicates that the next fields are other themes and should therefore be handled as such (this matters in the `apply_optional` function that the macro generates).

Example of using a theme (in `freya-components` crate):

```rust
use crate::theme::get_theme;
use freya_hooks::ButtonThemeWith;

#[derive(Props)]
pub struct ButtonProps {
    /// Theme override.
    #[props(optional)]
    pub theme: Option<ButtonThemeWith>,
    // ...
}

pub fn Button(cx: Scope<ButtonProps>) -> Element {
    // First argument is self explanatory, the second one is the `ThemeWith` struct, the third one is the name of the theme in the `Theme` struct
    let ButtonTheme { 
        padding,
        width,
        background,
        ..
    } = get_theme!(cx, &cx.props.theme, button);
    // ...
}
```

# Notes

This can be merged, but there is a couple things left to do that are relevant:

- [x] Move layout styles to themes.
- [ ] Eliminate repetition. This is the `theme` property definition, the `let theme = ...` statement, and the `/// This inherits [FooTheme](freya_hooks::FooTheme)` doc. It was extremely annoying adding all of these.
- [x] Make a macro for easier theme setting (get rid of `..Default::default` and `Some`). Update: see `theme_with` macro.
- [x] Document new system in book.

~- Move theme defs to component files.~ This is getting put on hold for now, since there's some cyclic dependency issues. However, in the future, it would indeed be good to move the themes somewhere else than `freya-hooks`.


tigerros:
There might be a bit of a problem with moving the themes to the component files... It's impossible because `Theme` is in the `hooks` crate, but it would depend on the themes in the `components` crate, but `components` depends on `hooks`. It's a cyclic dependency. `Theme` could also be relocated to `components`, but I don't know about all the theme hooks.

marc2332:
> There might be a bit of a problem with moving the themes to the component files... It's impossible because `Theme` is in the `hooks` crate, but it would depend on the themes in the `components` crate, but `components` depends on `hooks`. It's a cyclic dependency. `Theme` could also be relocated to `components`, but I don't know about all the theme hooks.

Heh, that explains why the themes are in the hooks crate. I would leave them in the hooks crate for now, maybe they can be moved to their own crate at some point in the future

tigerros:
That was painful, but it's done! All layout props were moved to the theme. All examples have been updated. Please mark this as breaking though.

tigerros:
Tried to update the book, but I probably didn't cover everything. As a side note, please check out [#427](https://github.com/marc2332/freya/issues/427). I tested the book to see any errors, but there was way too many of them and I couldn't be bothered to look through them to find out which ones are affected by this PR. [#427](https://github.com/marc2332/freya/issues/427) would fix this by making sure that the book has no errors.

tigerros:
Ready to be merged, but it's breaking, because layout stuff has been moved to themes.

PR #426, opened 2023-12-09
book: Fix mistake in theming chapter

tigerros:
The book says that the default theme is `DARK_THEME`, but it's actually `LIGHT_THEME`.


marc2332:
I forgot to update it when I changed it from dark to light ! ooops! ty!

PR #429, opened 2023-12-14
feat: Versioned docs

marc2332:
Move all the sections from the book to the `freya` crate that should be versioned and bring back the 0.1.8 book until 0.2.0 is released


tigerros:
Could you add the element property docs directly to the consts? E.g. move `opacity.md` to:

```
builder_constructors! {
  rect {
    /// Docs
    opacity: String,
    // ...
  }
}
```

The only change necessary is to modify the `impl_element` to include attributes for the const property.

tigerros:
Oh you already did that. Sorry 😅 I just saw the markdown files and thought you made them separate :)

PR #431, opened 2023-12-17
Versioned docs: better warnings + fix some typos

tigerros:
Instead of using `> Message`, we should use `<div class="warning">Message</div>`. Comparison:

doc-warning-comparison

### DIFF
--- a/crates/freya/src/_docs/animating.rs
+++ b/crates/freya/src/_docs/animating.rs
@@ -4,7 +4,7 @@
 //!
 //! ### `use_animation`
 //!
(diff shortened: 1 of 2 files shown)


marc2332:
isn't there a way to do it with Markdown? the warning thing

tigerros:
No, I don't think so. I looked at the commonmark reference and they do have a blockquote, but rustdoc just styles it like that. The div is what rustdoc recommends instead (https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block).

marc2332:
ah nice 👀

PR #432, opened 2023-12-18
Fix error in password example

tigerros:
This forwards changes from https://github.com/marc2332/freya/commit/e37f66390671552235ab216eefeb62040cc846af to the password example.


marc2332:
My bad.. ty! Perhaps I could add a carco check just for examples in the CI...

tigerros:
I agree. I think testing the doctests would be good too.

PR #433, opened 2023-12-18
Fix typo in `rect` prop

tigerros:
Defined property `decoraton_color` -> `decoration_color`. This was causing an error in one of the examples. Note that this is technically a breaking change.


marc2332:
I think I messed it up in the versioned docs PR, so its fine I think , I'll check later, ty!

PR #434, opened 2023-12-18
Add example and doc tests to workflow

tigerros:
This PR includes examples when testing and adds doctests **only for the `freya` crate** (which tests the book right now). That's because there's 114 failing doctests in other crates which I wasn't feeling like fixing, but I think imports are the source of most errors, so it shouldn't be that hard. I can do it, but I wouldn't mind if someone else did it.

There were some errors which I fixed to make all new tests pass, and not all of them were just about importing stuff. That's why I strongly think this PR should be merged, along with fixing all errors in the other docs and testing those too.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,12 +38,14 @@ jobs:
       run: |
         export RUSTFLAGS="-Cinstrument-coverage"
         export LLVM_PROFILE_FILE='cargo-test-%p-%m.profraw'
-        cargo nextest run --workspace
+        cargo nextest run --workspace --exclude examples
+        cargo check --examples
         cargo test -p freya --doc
     - name: Run MacOS and Windows tests
       if: runner.os != 'Linux'
       run: |
-        cargo nextest run --workspace
+        cargo nextest run --workspace --exclude examples
+        cargo check --examples
         cargo test -p freya --doc
     - name: Run coverage
       if: runner.os == 'Linux'
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,10 +38,13 @@ jobs:
       run: |
         export RUSTFLAGS="-Cinstrument-coverage"
         export LLVM_PROFILE_FILE='cargo-test-%p-%m.profraw'
-        cargo nextest run --workspace --exclude examples
+        cargo nextest run --workspace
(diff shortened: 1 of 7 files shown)


tigerros:
Seems like the Linux workflow doesn't have enough space. We could remove nextest, the cache (not quite sure that would help), or maybe get rid of some of the Linux dependencies (if possible).

.github/workflows/rust.yml
```diff
@@ -38,10 +38,13 @@ jobs:
       run: |
         export RUSTFLAGS="-Cinstrument-coverage"
         export LLVM_PROFILE_FILE='cargo-test-%p-%m.profraw'
-        cargo nextest run --workspace --exclude examples
+        cargo nextest run --workspace
```
marc2332, line 41: We need to bring back the `--exclude examples` because creating the binaries for the examples uses a crazy amount of storage, specially on Windows.

I would instead add a new command just for examples, something like this

```
cargo check --examples
```

marc2332 (changes requested):

.github/workflows/rust.yml
tigerros: added

crates/freya/src/_docs/hot_reload.rs
```diff
@@ -7,14 +7,18 @@
 //! Before launching your app, you need to initialize the hot-reload context:
 //!
 //! ```rust, no_run
-//! use freya::prelude::*;
-//! use freya::hot_reload::FreyaCtx;
+//! # use freya::prelude::*;
```
marc2332, line 10: I would actually leave this one visible, it makes more sense to me
```suggestion
//! use freya::prelude::*;
```

marc2332 (changes requested):

marc2332:
> Seems like the Linux workflow doesn't have enough space. We could remove nextest, the cache (not quite sure that would help), or maybe get rid of some of the Linux dependencies (if possible).

There are currently around 70 examples, and 1 single example compiled in debug weights like 300mb on Linux, I think. (I might completely wrong, I am on Windows rn)

70examples * 300MB = 21GB

That is the reason I excluded the examples

marc2332 (approved):
🔥

tigerros:
umm... why is this not merged?

PR #444, opened 2023-12-29
Improve `OrderDirection` docs

tigerros:
I was a bit surprised to see "Down" and "Up" for order rather than "Ascending" and "Descending", so this fixes that (doesn't rename the variants, just adds docs). Also adds a note that the table `order_direction` prop is only visual.

### DIFF
--- a/crates/components/src/table.rs
+++ b/crates/components/src/table.rs
@@ -110,9 +110,9 @@ pub fn TableRow<'a>(cx: Scope<'a, TableRowProps<'a>>) -> Element {
 
 #[derive(Clone, Copy, PartialEq, Debug, Default)]
 pub enum OrderDirection {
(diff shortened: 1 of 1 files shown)


marc2332:
I'm fine renaming these to `Ascending` / `Descending`, Up and down sounds weird here 😂

tigerros:
> I'm fine renaming these to `Ascending` / `Descending`, Up and down sounds weird here 😂

I just didn't want to introduce any breaking changes, that's all. Also, `Up` / `Down` don't sound weird when you consider the fact that it's only a visual change, but I suppose that may change in the future. I don't quite care though, which is I chose the non-breaking option.

marc2332:
np, I'll change it myself in another PR then

PR #452, opened 2024-01-01
docs: Improve docs and fix errors in doctests

tigerros:
This "builds on" https://github.com/marc2332/freya/pull/434. That PR fixed only the doctests (Rust code snippets) in the main crate docs though. This PR fixes every single code snippet and makes `cargo test --workspace --doc` pass. This command is also added to the workflow.

There's cca. 130 "new" (fixed and passing) doctests. All of them needed imports, but a lot of them didn't compile for other reasons. Other times, they compiled, but they were wrong in a different way (e.g. `max_lines.md` had a code snippet that was using `line_height`).

Additionally, the `freya_elements` docs were expanded in some cases (remove ambiguity, specify details), standardized (e.g. all examples are grouped under an example heading, previously some examples had just a plaintext `Example:` "title"), and fixed (link fixes, typos, etc.).

### DIFF
--- a/crates/elements/src/_docs/attributes/font_family.md
+++ b/crates/elements/src/_docs/attributes/font_family.md
@@ -1,10 +1,11 @@
-### font_family
+With the `font_family` you can specify what font you want to use for the inner text.
 
-With the `font_family` you can specify what font do you want to use for the inner text.
+Check out the [custom font example](https://github.com/marc2332/freya/blob/main/examples/custom_font.rs)
+to see how you can load your own fonts.
 
-Limitation: Only fonts installed in the system are supported for now.
+<!-- TODO: Example of checking if a font exists with skia_safe -->
 
-Example:
+### Example
 
 ```rust, no_run
 # use dioxus::prelude::*;
--- a/crates/elements/src/_docs/attributes/font_family.md
+++ b/crates/elements/src/_docs/attributes/font_family.md
@@ -8,8 +8,7 @@ to see how you can load your own fonts.
 ### Example
 
 ```rust, no_run
-# use dioxus::prelude::*;
-# use freya_elements::elements as dioxus_elements;
+# use freya::prelude::*;
 fn app(cx: Scope) -> Element {
     render!(
         label {
--- a/crates/elements/src/_docs/attributes/font_family.md
+++ b/crates/elements/src/_docs/attributes/font_family.md
@@ -7,6 +7,8 @@ Limitation: Only fonts installed in the system are supported for now.
 Example:
 
 ```rust, no_run
+# use dioxus::prelude::*;
+# use freya_elements::elements as dioxus_elements;
 fn app(cx: Scope) -> Element {
     render!(
         label {
--- a/crates/freya/src/lib.rs
+++ b/crates/freya/src/lib.rs
@@ -10,7 +10,7 @@
 //!
 //! - [Elements API reference](freya_elements::elements#structs)
 //! - [Events API reference](freya_elements::elements#functions)
-//! - [Element guides](freya_elements::_docs)
+//! - [Elements guides](freya_elements::_docs)
 //! - [Components](freya_components)
 //! - [Hooks](freya_hooks)
 //! - [Theming](self::_docs::theming)
--- a/crates/freya/src/lib.rs
+++ b/crates/freya/src/lib.rs
@@ -8,8 +8,8 @@
 //!
 //! Powered by [🧬 Dioxus](https://dioxuslabs.com) and [🎨 Skia](https://skia.org/).
 //!
-//! - [Elements](freya_elements::elements#structs)
-//! - [Events](freya_elements::elements#functions)
+//! - [Elements API reference](freya_elements::elements#structs)
+//! - [Events API reference](freya_elements::elements#functions)
 //! - [Elements and events guides](freya_elements::_docs)
 //! - [Components](freya_components)
 //! - [Hooks](freya_hooks)
--- a/crates/freya/src/lib.rs
+++ b/crates/freya/src/lib.rs
@@ -8,15 +8,16 @@
 //!
 //! Powered by [🧬 Dioxus](https://dioxuslabs.com) and [🎨 Skia](https://skia.org/).
 //!
-//! - [`Elements`](freya_elements::elements#structs)
-//! - [`Events`](freya_elements::elements#functions)
-//! - [`Components`](freya_components)
-//! - [`Hooks`](freya_hooks)
-//! - [`Theming`](self::_docs::theming)
-//! - [`Hot reload`](self::_docs::hot_reload)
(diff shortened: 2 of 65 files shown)


marc2332 (changes requested):
Instead of 

```
# use dioxus::prelude::*;
# use freya_elements::elements as dioxus_elements;
```

you could just do
```
# use freya::prelude::*;
```

For that you will need to add freya as a dev dep instead of Dioxus

tigerros:
> Instead of
> 
> ```
> # use dioxus::prelude::*;
> # use freya_elements::elements as dioxus_elements;
> ```
> 
> you could just do
> 
> ```
> # use freya::prelude::*;
> ```
> 
> For that you will need to add freya as a dev dep instead of Dioxus

Ah, true.

tigerros:
Actually, I can't. `render!` produces code with `::dioxus` in it.

marc2332:
> Actually, I can't. `render!` produces code with `::dioxus` in it.

True. Add back dioxus as a dep, but keep the code to `# use freya::prelude::*;` and that's it

tigerros:
> > Actually, I can't. `render!` produces code with `::dioxus` in it.
> 
> True. Add back dioxus as a dep, but keep the code to `# use freya::prelude::*;` and that's it

Didn't think of that. Should be good now!

crates/elements/src/_docs/attributes/font_family.md
```diff
@@ -1,12 +1,14 @@
-### font_family
+With the `font_family` you can specify what font you want to use for the inner text.
 
-With the `font_family` you can specify what font do you want to use for the inner text.
+Check out the [custom font example](https://github.com/marc2332/freya/blob/main/examples/custom_font.rs)
+to see how you can load your own fonts.
 
-Limitation: Only fonts installed in the system are supported for now.
+<!-- TODO: Example of checking if a font exists with skia_safe -->
```
marc2332, line 6: > TODO: Example of checking if a font exists with skia_safe

What do you mean by this?

crates/freya/src/lib.rs
```diff
@@ -8,15 +8,16 @@
...
-//! - [`Components`](freya_components)
-//! - [`Hooks`](freya_hooks)
-//! - [`Theming`](self::_docs::theming)
-//! - [`Hot reload`](self::_docs::hot_reload)
-//! - [`Testing`](self::_docs::testing)
-//! - [`Animating`](self::_docs::animating)
-//! - [`Devtools`](self::_docs::devtools)
+//! - [Elements](freya_elements::elements#structs)
+//! - [Events](freya_elements::elements#functions)
+//! - [Elements and events guides](freya_elements::_docs)
```
marc2332, line 13: > //! - [Elements and events guides](freya_elements::_docs)

This working seems odd to me. There is no event mention of events in those docs, and leaving `Elements guides` also sounds weird to me, isn't there maybe a better way to put this?

marc2332 (changes requested):
just this, the rest LGTM

crates/elements/src/_docs/attributes/font_family.md
tigerros: I removed the "limitation" because it's not true since you can load your own fonts.

As for the todo, I thought it makes sense to include a way to check if a font exists, rather than just "hope" that it exists. Like if you use "Inter" as a font, you might want to know if it exists, and fallback to another font.

marc2332: A priority list for `font_family` could be implemented, just like in CSS, what do you think?

crates/freya/src/lib.rs
tigerros: I used "elements and events" because I assume it might have some event docs in the future. It can be "Elements guides", but if we forget to change it once some event "guide" is added there, people might just assume there's no additional event docs other than the API.

As for the ".. guides" name, I don't really see the problem. That's what they're referred to on the website, so I just used that.

marc2332: Well, what triggers me off is that there is elements, events, and then elements & events again, reads weird to me

crates/elements/src/_docs/attributes/font_family.md
tigerros: Do you mean Serif, Sans-serif, Monospace, Cursive and Fantasy? That would be nice yeah.

crates/freya/src/lib.rs
tigerros: I changed "Elements" and "Events" to "Elements API reference", "Events API reference". Should be more clear.

marc2332: that's weird considering all the other things from the list are also API references xD

tigerros: The other things on the list don't have "guides" though.

tigerros:
Something's wrong with the ubuntu workflow but I don't know what it is. It just gets stuck at a certain point. Please help

marc2332:
> Something's wrong with the ubuntu workflow but I don't know what it is. It just gets stuck at a certain point. Please help

probably a doc test calling a launch function

marc2332:
So weird, it runs just fine in my Windows and PopOS. And I can't seem to find the reason the compilation fails, cargo doesn't say it

tigerros:
I put the doctests and examples in separate jobs. Works now.

marc2332:
> I put the doctests and examples in separate jobs. Works now.

Nice! That was realllly weird, never seen any CI doing that

crates/freya/src/lib.rs
```diff
@@ -8,15 +8,16 @@
...
-//! - [`Components`](freya_components)
-//! - [`Hooks`](freya_hooks)
-//! - [`Theming`](self::_docs::theming)
-//! - [`Hot reload`](self::_docs::hot_reload)
-//! - [`Testing`](self::_docs::testing)
-//! - [`Animating`](self::_docs::animating)
-//! - [`Devtools`](self::_docs::devtools)
+//! - [Elements API reference](freya_elements::elements#structs)
+//! - [Events API reference](freya_elements::elements#functions)
+//! - [Elements and events guides](freya_elements::_docs)
```
marc2332, line 13: What about naming this to:
```suggestion
//! - [Style and layout explanations](freya_elements::_docs)
```

marc2332: or "references" instead of "explanations"

marc2332 (changes requested):

crates/freya/src/lib.rs
marc2332: > "Style and layout" is misleading because the inheritance "guide" is neither of those. I know there's nothing there about events, but I believe there will be in the future. 

What about `Supported colors, layout and inherited values`, also, if there are no events, let's not call it events, this are versioned docs, if there is ever some events explanation this can be changed.

marc2332 (approved):

PR #466, opened 2024-01-17
Remove required trailing comma in `theme_with` macro

tigerros:
I tend to make trailing commas required for standardization but public macros should be more permissive especially since the error without the trailing comma is not very self explanatory.


marc2332:
Thanks!

PR #467, opened 2024-01-20
Remove outdated info from theming docs

tigerros:


crates/freya/src/_docs/theming.rs
```diff
@@ -2,8 +2,6 @@
 //!
 //! Freya has built-in support for Theming.
 //!
-//! <div class="warning">⚠️ As of 2023-12-19, extending the base theme is not supported.</div>
```
marc2332, line 5: Isn't this still correct? You can't add your own stuff to a theme

marc2332 (changes requested):

crates/freya/src/_docs/theming.rs
marc2332: Yeah, that's it, it means that you cannot add any custom fields to an instance of the `Theme` struct

marc2332: > > Yeah, that's it, it means that you cannot add any custom fields to an instance of the `Theme` struct
> 
> You can make your own themes though. Not being able to add fields to the theme struct will always be impossible anyway, unless we make it a trait.

You never know what can happen in the future :)

marc2332: Let's add the note back!

marc2332: > Sorry, I'm in the hospital. Can't do much.

Damn, no worries. Take care!

PR #521, opened 2024-02-24
feat: Faster measurement when using alignments

marc2332:


marc2332:
I should probably add benchmarks for multiple nested elements

PR #532, opened 2024-03-01
feat: Cache `SkParagraph` between frames

marc2332:
I have only noticed very small improvements in my Windows, perhaps it will have a better impact on lower end devices.


marc2332:
Have yet to update tests, benchmarks and examples of torin accordingly.

PR #533, opened 2024-03-02
feat: Add `use_asset_cacher`

marc2332:
Closes https://github.com/marc2332/freya/issues/520


crates/components/src/network_image.rs
```diff
@@ -57,33 +62,52 @@ pub enum ImageStatus {
...
-                image_bytes.set(Some(img));
-                status.set(ImageStatus::Loaded)
-            } else if let Err(_err) = img {
-                // Image errored
-                image_bytes.set(None);
-                status.set(ImageStatus::Errored)
-            }
-        });
+        let asset_configuration = AssetConfiguration {
+            duration: Duration::from_secs(3600),
```
marc2332, line 76: I should probably make the duration a prop with a better default or something

marc2332 (commented):
~a precache method would be useful~

I actually already have that

marc2332:
Blocked by https://github.com/DioxusLabs/dioxus/issues/2009

Issue #482, opened 2024-01-27
task: Fork Dioxus `native-core` crates


marc2332:
There is a high change that Dioxus will deprecrate [Native Core](https://github.com/DioxusLabs/dioxus/tree/master/packages/native-core) and [Native core macro](https://github.com/DioxusLabs/dioxus/tree/master/packages/native-core-macro), and since Freya depends on these two, I have thought about forking. The Dioxus maintainers are OK with it too

This might open some optimizations I was not able to do before, I'll see

Some optimizations that come to my mind:
- [ ] Remove the utils folders
- [ ] Use Enums to store the `tag`, `attributes` and `listeners` in `ElementNode` and other String fields in `OwnedAttributeDiscription
- [ ] Remove `namespace` where possible
- [ ] Remove generics in favor of using the direct types used in Freya
- [ ] Remove Custom Elements (?)
- [ ] Remove layout_attributes (?)
- [ ] Not storing Node states for Text nodes as  they are not used in freya
- [ ] ~Simplify `get_listening_sorted`~
- [ ] Remove listeners from text nodes
    -  Tips from Evan:
       - https://github.com/DioxusLabs/dioxus/blob/c518c91c4c001639a24f8d88f4e1c7c78dc2eff6/packages/native-core/src/dioxus.rs#L116-L120
       - https://github.com/DioxusLabs/dioxus/blob/master/packages/native-core/src/real_dom.rs#L315-L384

felipetesc:
I think this idea will bring more autonomy, and allow Freya to follow her own path, but it would also add another layer of complexity. So far, using Dioxus allows you not to worry about a series of problems, but as you correctly observed, it makes it difficult to improve performance.
Programming user interface libraries is not easy and having more control over the code is a way to improve the project. I think the Rust community will greatly benefit from a library without dependency hell issues.

marc2332:
> I think this idea will bring more autonomy, and allow Freya to follow her own path, but it would also add another layer of complexity. So far, using Dioxus allows you not to worry about a series of problems, but as you correctly observed, it makes it difficult to improve performance. Programming user interface libraries is not easy and having more control over the code is a way to improve the project. I think the Rust community will greatly benefit from a library without dependency hell issues.

Instead of "another layer of complexity" I would say "more code to maintain". But yeah, it is true that not having to worry about it would be certainly better when it comes to maintenance, but I think the potential performance optimizations and perhaps removal of abstractions in favor of implicit implementations might be worth doing

[...]

PR #537, opened 2024-03-16
feat: Simplify and adapt `native core`

marc2332:
Depends on https://github.com/marc2332/freya/pull/535
Closes https://github.com/marc2332/freya/issues/482

- [x] Simplify `TextNode`
- [x] Use Enums to store the `tag`, `attributes` and `listeners` in `ElementNode` and other String fields in `OwnedAttributeDiscription`
- [x] Remove `namespace` where possible
- [x] Remove Custom Elements (?)
- [x] Not storing Node states for Text nodes as they are not used in freya
- [x] Remove listeners from text nodes
    -  Tips from Evan:
       - https://github.com/DioxusLabs/dioxus/blob/c518c91c4c001639a24f8d88f4e1c7c78dc2eff6/packages/native-core/src/dioxus.rs#L116-L120
       - https://github.com/DioxusLabs/dioxus/blob/master/packages/native-core/src/real_dom.rs#L315-L384


marc2332:
Devtools seem to be broken

PR #625, opened 2024-05-27
fix: `WindowDragArea` component hanging on non-left click.

dload0:
When the `WindowDragArea` component received a non-left click, the drag area would become disabled, even when receiving standard right click events.

Fixed by adding a check in the `onmousedown` function within the component.


marc2332:
Thanks!

Issue #560, opened 2024-03-31
enhancement: Use System Fonts


marc2332:
Use the OS default font instead of Fira Sans

PR #661, opened 2024-06-06
feat: Use System fonts

marc2332:
Closes https://github.com/marc2332/freya/issues/560


crates/core/src/style.rs
```diff
@@ -0,0 +1,12 @@
+pub fn default_fonts() -> Vec<String> {
+    let mut fonts = vec!["Noto Sans".to_string(), "Arial".to_string()];
+    if cfg!(target_os = "windows") {
+        fonts.insert(0, "Segoe UI".to_string());
+        fonts.insert(1, "Segoe UI Emoji".to_string());
+    } else if cfg!(target_os = "macos") {
+        fonts.insert(0, "San Francisco (SF)".to_string());
```
marc2332, line 7:
```suggestion
        fonts.insert(0, ".AppleSystemUIFont".to_string());
```

PR #664, opened 2024-06-08
feat: Optionally allow inserting tabs as spaces

marc2332:
Closes https://github.com/marc2332/freya/issues/662


crates/hooks/src/rope_editor.rs
```diff
@@ -10,6 +10,7 @@ use crate::{text_editor::*, EditableMode, EditorHistory, HistoryChange};
 pub struct RopeEditor {
     pub(crate) rope: Rope,
     pub(crate) cursor: TextCursor,
+    pub(crate) identation: usize,
```
marc2332, line 13: should probably use u8

Issue #602, opened 2024-05-04
bug: Replaced elements from loaded from a batch in the same frame leads to a crash


ZeroX-DG:
I encounter this error when I resize my app really quick. It doesn't happen on freya examples so I'm not sure what is causing this.

```
thread 'main' panicked at ~/.cargo/git/checkouts/freya-aa88117f1f713ee2/7b68806/crates/core/src/dom/mutations_writer.rs:46:60:
called `Option::unwrap()` on a `None` value
thread 'main' panicked at ~/.cargo/registry/src/index.crates.io-6f17d22bba15001f/winit-0.29.15/src/platform_impl/macos/app_state.rs:387:33:
called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```

marc2332:
Looks like some kind of edge case I missed, will look into it asap!

marc2332:
hey, I can't reproduce it, so I pushed a branch with some logs which could help me debug it -> https://github.com/marc2332/freya/tree/fix/mutations-writer-error

Could you run your app with it and share me the logs once it crashes?

marc2332:
Would also appreciate it if you could narrow it down in your app

ZeroX-DG:
Hey, here's the resulting log:

```
EId(58.0) Element(ElementNode { tag: Rect, attributes: {Layer: Text("-10"), Width: Text("8.4"), Background: Text("rgb(165, 172, 186)"), PositionTop: Text("0"), Height: Text("18"), Color: Text("rgb(17, 21, 28)"), PositionLeft: Text("42"), Position: Text("absolute")}, listeners: {} }) 8
EId(59.0) Element(ElementNode { tag: Rect, attributes: {}, listeners: {} }) 9
EId(60.0) Element(ElementNode { tag: Label, attributes: {}, listeners: {} }) 10
EId(60.1) Element(ElementNode { tag: Rect, attributes: {Layer: Text("-10"), Width: Text("8.4"), Background: Text("rgb(165, 172, 186)"), PositionTop: Text("0"), Height: Text("18"), Color: Text("rgb(17, 21, 28)"), PositionLeft: Text("8.4"), Position: Text("absolute")}, listeners: {} }) 8
EId(59.1) Element(ElementNode { tag: Rect, attributes: {}, listeners: {} }) 9
EId(58.1) Element(ElementNode { tag: Label, attributes: {}, listeners: {} }) 10
[...]

PR #744, opened 2024-06-27
fix: Skip DOM Nodes loaded in the same mutations run

marc2332:
Closes https://github.com/marc2332/freya/issues/602

### DIFF
--- a/crates/core/src/dom/mutations_writer.rs
+++ b/crates/core/src/dom/mutations_writer.rs
@@ -49,13 +49,12 @@ impl<'a> MutationsWriter<'a> {
                 }
 
                 let layer_state = node.get::<LayerState>();
-                let does_node_have_state = layer_state.is_some();
 
                 // There might exist Nodes in the RealDOM with no states yet,
                 // this is mainly due to nodes being created in the same run as when this function (remove) is being called,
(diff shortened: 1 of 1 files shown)


crates/core/src/dom/mutations_writer.rs
```diff
@@ -48,6 +48,17 @@ impl<'a> MutationsWriter<'a> {
...
                 }
 
+                let layer_state = node.get::<LayerState>();
+                let does_node_have_state = layer_state.is_some();
+
+                // There might exist Nodes in the RealDOM with no states yet,
+                // this is mainly due to nodes being created in the same run as when this function (remove) is being called,
+                // like nodes created by loaded templates.
+                // In this case we can safely skip these nodes.
+                if !does_node_have_state {
```
ZeroX-DG, line 58: Nitpicking: Maybe just call the variable `node_state_is_empty = layer_state.is_none()` so we don't have to think about the `!` haha.

ZeroX-DG:
Looks nice. I'll give it a quick test when I'm free haha.

ZeroX-DG:
Works great for me!

ZeroX-DG (approved):
LGTM 🎉

PR #754, opened 2024-06-30
Components preview POC

anvlkv:


examples/all-components/src/ds/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+            cross_align: "center",
+            height: "100%",
+            width: "100%",
+            rect {
+                height: "80%",
+                width: "100%",
+                ThemeProvider {
+                    theme: theme_signal()
+                    {children}
+                }
```
marc2332, line 27: `ThemeProvider` already calls `use_init_default_theme` internally, we don't need it here.
```suggestion
                {children}
```

examples/all-components/src/ds/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+            }
+            rect{
+                height: "20%",
+                width: "100%",
+                ThemeEditor{
+                    theme: theme_signal
```
marc2332, line 33: You don't really need to pass the theme signal as prop, instead you can just call `use_theme` in all the components you need the signal.

marc2332 (changes requested):

marc2332:
Closing to inactivity

PR #760, opened 2024-07-02
feat: Replace `mouseover` with `mousemove` and add new `mouseover` and `mouseout` events

marc2332:
- Replaces `mouseover` with `mousemove`
- Adds a new `mouseover`, similar to `mousenter` but only for the actual visible area instead of also triggering uppon entering child elements.
- Adds a new `mouseout` to be used in combination with `mouseover`, so just like the former one it would be similar to `mouseleave`, but only trigger when leaving the visible area, which might include hovering a child element for example.

## To Do
- Examples
- Tests
- Docs
- check if `does_bubble` is correct

### DIFF
--- a/crates/core/tests/pointer_events.rs
+++ b/crates/core/tests/pointer_events.rs
@@ -12,7 +12,7 @@ pub async fn pointer_events_from_mouse() {
 
         let onpointerup = move |_| state.push("up".to_string());
 
-        let onpointerover = move |_| state.push("over".to_string());
+        let onpointermove = move |_| state.push("over".to_string());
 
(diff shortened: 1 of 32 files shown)


crates/core/tests/pointer_events.rs
```diff
@@ -120,7 +120,7 @@ pub async fn pointer_events_from_touch() {
 
         let onpointerup = move |_| state.push("up".to_string());
 
-        let onpointerover = move |_| state.push("over".to_string());
+        let onpointermove = move |_| state.push("over".to_string());
```
marc2332, line 123:
```suggestion
        let onpointermove = move |_| state.push("move".to_string());
```

Issue #762, opened 2024-07-04
bug: Opacity attribute prevents shadows rendering correctly.


dload0:
When having a similar setup to the code below, the shadow in the child element will not render.

```
rect {
    height: "100%",
    width: "100%",
    main_align: "center",
    cross_align: "center",

    opacity: "1", // <- `opacity` attribute in parent element will prevent shadows in child elements from being rendered.

    rect {
        width: "50",
        height: "50",
        background: "black",

        shadow: "0 0 10 5 rgb(0, 0, 0)",
    }
}
```

As well as this, applying an opacity attribute to the element itself with the shadow will have the same behaviour.

```
rect {
    width: "50",
    height: "50",
    background: "black",
    
    opacity: "1", // Will still render the component, however the shadow will not render.

    shadow: "0 0 10 5 rgb(0, 0, 0)",
}
```

PR #764, opened 2024-07-06
fix: Prevent opacity from clipping the node bounds

marc2332:
Closes https://github.com/marc2332/freya/issues/762


marc2332:
idea: Prepare shadows, adjust node bounds with the shadows bounds and then apply the opacity on the adjusted bounds

Issue #767, opened 2024-07-07
bug: The `overflow` attribute does not take into account the value of the `corner_radius` attribute.


Aiving:
I have recently started using the library and am generally very satisfied! However, while trying to implement the ripple effect from Material Design, I came across the fact that the `overflow` attribute ignores the value of the `corner_radius` attribute. Regardless of its value, clipping is always done using a simple rectangle. 

Sample code showing this bug:
```rs
use freya::prelude::*;

fn main() {
    launch(app);
}

fn app() -> Element {
    rsx! {
        rect {
            width: "fill",
            height: "fill",
            main_align: "center",
            cross_align: "center",
            
            rect {
                width: "200",
                height: "200",
                corner_radius: "12",
                overflow: "clip",

                // Should be clipped by a rounded rectangle
                rect {
                    width: "fill",
                    height: "fill",
                    background: "green"
                }
            }
        }
    }
}
```

marc2332:
Already have an idea on how to fix it, will open a PR soon

Aiving:
Good to know!

marc2332:
![image](https://github.com/marc2332/freya/assets/38158676/571d081d-4118-4958-b964-a8a765318f36)

Already got it working, just need to clean up the code 😄

marc2332:
Got a PR on draft here https://github.com/marc2332/freya/pull/768 in case you want to give it a try @Aiving

PR #768, opened 2024-07-07
fix: Consider corner radius for events and overflow clipping

marc2332:
Closes https://github.com/marc2332/freya/issues/767


marc2332:
I am still not 100% sold in the implementation because it is only applying the viewport rendering visually, so, events will still ignore the corner radius. I might have to rethink the viewports general idea

marc2332:
> I am still not 100% sold in the implementation because it is only applying the viewport when rendering visually, so, events will still ignore the corner radius. I might have to rethink the viewports general idea

This should be easy to do. Will try it asap

marc2332:
> > I am still not 100% sold in the implementation because it is only applying the viewport when rendering visually, so, events will still ignore the corner radius. I might have to rethink the viewports general idea
> 
> This should be easy to do. Will try it asap

done!

Issue #770, opened 2024-07-08
enhancement: Support `alpha` channel in `AnimColor`


Aiving:
The current implementation of `AnimColor` returns (and animates) only RGB when calling `as_string`, even if in `origin` or `destination` properties the `alpha` channel value is not 255.

https://github.com/marc2332/freya/blob/1955629e6c355ff74415037e44869026da8aec89/crates/hooks/src/use_animation.rs#L180-L187

marc2332:
Very good point. I'll improve this asap!

marc2332:
Just opened https://github.com/marc2332/freya/pull/771

PR #771, opened 2024-07-08
fix: Rethink rendering of elements

marc2332:
Closes https://github.com/marc2332/freya/issues/770
Depends on https://github.com/marc2332/freya/pull/768

### DIFF
--- a/crates/hooks/src/use_animation.rs
+++ b/crates/hooks/src/use_animation.rs
@@ -179,10 +179,11 @@ impl AnimatedValue for AnimColor {
 
     fn as_string(&self) -> String {
         format!(
-            "rgb({}, {}, {})",
+            "rgb({}, {}, {}, {})",
             self.value.r(),
             self.value.g(),
-            self.value.b()
+            self.value.b(),
+            self.value.a()
         )
     }
 


Aiving:
Despite the support of the alpha channel in `as_string`, it still does not animate, since the `is_finished` and `advance` functions do not use or change the alpha channel

https://github.com/marc2332/freya/blob/fe3f4d35af1d6c2a3b0b5756494096c176ea5f3d/crates/hooks/src/use_animation.rs#L199-L249

marc2332:
> Despite the support of the alpha channel in `as_string`, it still does not animate, since the `is_finished` and `advance` functions do not use or change the alpha channel
> 
> https://github.com/marc2332/freya/blob/fe3f4d35af1d6c2a3b0b5756494096c176ea5f3d/crates/hooks/src/use_animation.rs#L199-L249

oh man, let me see if that can be done

marc2332:
> Despite the support of the alpha channel in `as_string`, it still does not animate, since the `is_finished` and `advance` functions do not use or change the alpha channel
> 
> https://github.com/marc2332/freya/blob/fe3f4d35af1d6c2a3b0b5756494096c176ea5f3d/crates/hooks/src/use_animation.rs#L199-L249

Fixed, I think

PR #785, opened 2024-07-13
feat: Rethink parsing

Aiving:


crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn as_str(&self) -> &str {
+        if let Token::Ident(value) = self {
+            value.as_str()
+        } else {
+            unreachable!()
+        }
+    }
+
+    pub fn as_string(&self) -> Option<&str> {
```
marc2332, line 75:
```suggestion
    pub fn try_as_str(&self) -> Option<&str> {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+impl Token {
+    pub fn ident<T: Into<String>>(value: T) -> Self {
+        Self::Ident(value.into())
+    }
+
+    pub fn is_ident(&self) -> bool {
+        matches!(self, Token::Ident(_))
+    }
+
+    pub fn is_float(&self) -> bool {
```
marc2332, line 29:
```suggestion
    pub fn is_f32(&self) -> bool {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+        matches!(self, Token::Float(_))
+    }
+
+    pub fn is_number(&self) -> bool {
```
marc2332, line 33:
```suggestion
    pub fn is_i64(&self) -> bool {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+        matches!(self, Token::Number(_))
+    }
+
+    pub fn is_integer(&self) -> bool {
```
marc2332, line 37:
```suggestion
    pub fn is_i64_or_f32(&self) -> bool {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn as_string(&self) -> Option<&str> {
+        if let Token::Ident(value) = self {
+            Some(value.as_str())
+        } else {
+            None
+        }
+    }
+
+    pub fn as_float(&self) -> Option<f32> {
```
marc2332, line 83:
```suggestion
    pub fn try_as_f32(&self) -> Option<f32> {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+        if let Token::Float(value) = self {
+            Some(*value)
+        } else if let Token::Number(value) = self {
+            Some(*value as f32)
+        } else {
+            None
+        }
+    }
+
+    pub fn as_number(&self) -> Option<i64> {
```
marc2332, line 93:
```suggestion
    pub fn try_as_i64(&self) -> Option<i64> {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+        if let Token::Float(value) = self {
+            value
+        } else if let Token::Number(value) = self {
+            value as f32
+        } else {
+            unreachable!()
+        }
+    }
+
+    pub fn into_number(self) -> i64 {
```
marc2332, line 59:
```suggestion
    pub fn as_i64(self) -> i64 {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn into_string(self) -> String {
+        if let Token::Ident(value) = self {
+            value
+        } else {
+            unreachable!()
+        }
+    }
+
+    pub fn into_float(self) -> f32 {
```
marc2332, line 49:
```suggestion
    pub fn as_f32(self) -> f32 {
```

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    pub fn is_number(&self) -> bool {
+        matches!(self, Token::Number(_))
+    }
+
+    pub fn is_integer(&self) -> bool {
+        matches!(self, Token::Number(_) | Token::Float(_))
+    }
+
+    pub fn into_string(self) -> String {
```
marc2332, line 41:
```suggestion
    pub fn as_string(self) -> String {
```

marc2332 (changes requested):
Having to write 

```rs
let mut parser = Parser::new(Lexer::parse(value));
```
every time in the node states is a step back, why aren't those using `parse_value`?

Aiving:
> Having to write
> 
> ```rust
> let mut parser = Parser::new(Lexer::parse(value));
> ```
> 
> every time in the node states is a step back, why aren't those using `parse_value`?

I forgot to change the node states code when I added the parse_value, will make fix commit soon

Aiving:
@marc2332 done

crates/state/src/font_style.rs
```diff
@@ -105,28 +106,26 @@ impl ParseAttribute for FontStyleState {
...
                     }
                 }
             }
             AttributeName::TextShadow => {
                 if let Some(value) = attr.value.as_text() {
-                    self.text_shadows = value
-                        .split_excluding_group(',', '(', ')')
-                        .map(|chunk| TextShadow::parse(chunk).unwrap_or_default())
-                        .collect();
+                    self.text_shadows = TextShadow::parse_values(value, &Token::Comma)?;
```
marc2332, line 122: Wouldn't using ( or ) here not make a difference?

crates/state/src/values/color.rs
```diff
@@ -13,29 +15,40 @@ pub trait DisplayColor {
...
+                    "red" => Ok(Color::RED),
+                    "green" => Ok(Color::GREEN),
+                    "blue" => Ok(Color::BLUE),
+                    "yellow" => Ok(Color::YELLOW),
+                    "black" => Ok(Color::BLACK),
+                    "gray" => Ok(Color::GRAY),
+                    "white" => Ok(Color::WHITE),
+                    "orange" => Ok(Color::from_rgb(255, 165, 0)),
+                    "transparent" => Ok(Color::TRANSPARENT),
+                    _ => unreachable!(),
```
marc2332, line 48: Why `unreachable!()`?

crates/state/src/values/color.rs
```diff
@@ -72,127 +85,129 @@ impl DisplayColor for Color {
...
+    parser.consume(&Token::Comma)?;
+
+    let blue = parser.consume_map(parse_number_as)?;
+
+    let color = if parser.try_consume(&Token::Comma) {
+        let alpha = parser.consume_if(Token::is_i64_or_f32).and_then(|token| {
+            if token.is_i64() {
+                u8::try_from(token.into_i64()).map_err(|_| ParseError)
+            } else {
+                Ok((token.into_f32() * 255.0).round().clamp(0.0, 255.0) as u8)
```
marc2332, line 110: you could add a `into_u8` instead of making the conversion ere

crates/state/src/parsing.rs
```diff
@@ -1,15 +1,142 @@
...
+        }
+    }
+}
+
 // FromStr but we own it so we can impl it on torin and skia_safe types.
 pub trait Parse: Sized {
-    fn parse(value: &str) -> Result<Self, ParseError>;
+    fn parse(parser: &mut Parser) -> Result<Self, ParseError>;
+
+    fn parse_value(value: &str) -> Result<Self, ParseError> {
```
marc2332, line 113:
```suggestion
    fn parse(value: &str) -> Result<Self, ParseError> {
```

crates/state/src/parsing.rs
```diff
@@ -1,15 +1,142 @@
...
+        let value = Self::parse(&mut parser);
+
+        if parser.tokens.len() > 0 {
+            Err(ParseError)
+        } else {
+            value
+        }
+    }
+
+    fn parse_values(value: &str, separator: &Token) -> Result<Vec<Self>, ParseError> {
```
marc2332, line 125:
```suggestion
    fn parse_with_separator(value: &str, separator: &Token) -> Result<Vec<Self>, ParseError> {
```

crates/state/src/parsing.rs
```diff
@@ -1,15 +1,142 @@
...
+        } else {
+            None
+        }
+    }
+}
+
 // FromStr but we own it so we can impl it on torin and skia_safe types.
 pub trait Parse: Sized {
-    fn parse(value: &str) -> Result<Self, ParseError>;
+    fn parse(parser: &mut Parser) -> Result<Self, ParseError>;
```
marc2332, line 111:
```suggestion
    fn from_parser(parser: &mut Parser) -> Result<Self, ParseError>;
```

crates/state/src/parsing.rs
marc2332: This will make the API a bit more friendly and also bring down the amount of changed files in this PR

crates/state/src/values/color.rs
```diff
@@ -13,29 +15,40 @@ pub trait DisplayColor {
...
+                Token::ident("hsl"),
+                Token::ident("red"),
+                Token::ident("green"),
+                Token::ident("blue"),
+                Token::ident("yellow"),
+                Token::ident("black"),
+                Token::ident("gray"),
+                Token::ident("white"),
+                Token::ident("orange"),
+                Token::ident("transparent"),
```
marc2332, line 32: Is it possible to not have to specify all these values twice? Here and below in the match

marc2332 (changes requested):

crates/state/src/values/color.rs
Aiving: because `and_then` will be called only if the result is `Ok`, and `consume_one_of` will return `Err` if the current token is not among `values`

Aiving: Whoops, I mixed up the comments

Aiving: Regarding specifying all these values twice - I don't like it myself, but otherwise `&mut parser` can't be used for `parser_hex_color`, `parser_gb` and `parse_hsl` (because the parser will already be borrowed as mutable when `consume_map` is called)

Aiving: Ah, I figured out how to solve that problem.

crates/state/src/values/color.rs
Aiving: Instead of `into_u8', I think `try_as_u8' would be more appropriate.

marc2332: Ah yes 💯

crates/state/src/font_style.rs
Aiving: I don't quite understand how `(` or `)` would be used when parsing `text-shadow` (I haven't found an example of using `text-shadow` with them anywhere either)

crates/state/src/lexing.rs
```diff
@@ -0,0 +1,186 @@
+use std::iter;
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum Token {
+    Ident(String),
+    Float(f32),
+    Number(i64),
```
marc2332, line 7:
```suggestion
    Integer(i64),
```

marc2332 (changes requested):

crates/state/src/values/font.rs
```diff
@@ -3,46 +3,53 @@ use freya_engine::prelude::*;
...
-    fn parse(value: &str) -> Result<Self, ParseError> {
-        Ok(match value {
-            "upright" => Slant::Upright,
-            "italic" => Slant::Italic,
-            "oblique" => Slant::Oblique,
-            _ => Slant::Upright,
+    fn from_parser(parser: &mut Parser) -> Result<Self, ParseError> {
+        parser.consume_map(|token| {
+            token.try_as_str().and_then(|value| match value {
+                "upright" => Some(Slant::Upright),
```
marc2332, line 29:
```suggestion
                "upright" | "normal" => Some(Slant::Upright),
```

marc2332 (changes requested):

marc2332:
Can you make sure that all the examples run just fine @Aiving ?

marc2332:
> Can you make sure that all the examples run just fine @Aiving ?

You might do this by running the `examples.sh` script

Aiving:
> > Can you make sure that all the examples run just fine @Aiving ?
> 
> You may do this by running the `examples.sh` script

oki, i'll try

crates/state/src/values/font.rs
```diff
@@ -109,11 +124,13 @@ impl TextOverflow {
...
-            "clip" => TextOverflow::Clip,
-            value => TextOverflow::Custom(value.to_string()),
+    fn from_parser(parser: &mut Parser) -> Result<Self, ParseError> {
+        parser.consume_map(|token| {
+            token.try_as_str().map(|value| match value {
+                "ellipsis" => TextOverflow::Ellipsis,
+                "clip" => TextOverflow::Clip,
+                value => TextOverflow::Custom(value.to_string()),
+            })
         })
```
marc2332, line 134:
```suggestion
    fn from_parser(_parser: &mut Parser) -> Result<Self, ParseError> {
        Ok(match value {
            "ellipsis" => TextOverflow::Ellipsis,
            "clip" => TextOverflow::Clip,
            value => TextOverflow::Custom(value.to_string()),
        })
```

marc2332 (changes requested):

marc2332:
Conflicts @Aiving

Aiving:
> Conflicts @Aiving

resolved!

marc2332:
We need to make sure no example broke

marc2332:
Closing this, we can revisit this in the future, but for now I dont think adding this abstraction is really useful, I appreciate the effort @Aiving though, thanks :)

PR #789, opened 2024-07-14
Use Metal on macOS

reyamir:


crates/renderer/src/skia/metal_surface.rs
marc2332, line 1: Where is the surface? 🤔

crates/renderer/src/skia/gl_surface.rs
```diff
@@ -0,0 +1,176 @@
...
+        GlSurface, Surface as GlutinSurface, SurfaceAttributesBuilder, SwapInterval, WindowSurface,
+    },
+};
+use std::{ffi::CString, num::NonZeroU32};
+use winit::{raw_window_handle::HasWindowHandle, window::Window};
+use winit::dpi::PhysicalSize;
+
+use freya_engine::prelude::*;
+
+pub struct SkiaSurface {
```
marc2332, line 20: I would probably rename this to something like GlSurface and then have a type alias be stored in the window state instead, same thing for the metal surface `MetalSurface`

crates/renderer/src/skia/metal_surface.rs
```diff
@@ -0,0 +1,66 @@
...
+use objc::runtime::YES;
+use winit::{raw_window_handle::HasWindowHandle, window::Window};
+
+use freya_engine::prelude::*;
+
+pub struct SkiaSurface {
+    pub(crate) gr_context: DirectContext,
+    pub(crate) metal_layer: MetalLayer,
+    pub(crate) command_queue: CommandQueue,
+    pub(crate) scale_factor: f64,
```
marc2332, line 14: I don't think scale_factor should stored in the surface

marc2332 (changes requested):

reyamir:
I will rework on this on this weekend, as well as add vulkan support too.

marc2332:
hey @reyamir ! I just opened https://github.com/marc2332/freya/pull/822 which should make adding metal easier :)

reyamir:
> hey @reyamir ! I just opened #822 which should make adding metal and other drivers easier :)

that great, I will rewrite my PR

PR #790, opened 2024-07-14
feat: `import_svg` macro

marc2332:


crates/components/src/svg.rs
```diff
@@ -0,0 +1,17 @@
...
+    ($component_name:ident, $path:expr, $width: expr, $height: expr) => {
+        // Generate a function with the name derived from the file name
+        #[allow(non_snake_case)]
+        pub fn $component_name() -> freya::prelude::Element {
+            use freya::prelude::*;
+            let svg_content = include_str!($path);
+
+            rsx!(svg {
+                width: $width,
+                height: $height,
```
marc2332, line 12: should probably leave these two for props

PR #795, opened 2024-07-16
Add support for reverse ScrollView

reyamir:
How to use : just add `reverse: true` to ScrollView
Use case: chat app


marc2332 (changes requested):
- CI is not passing
- this only works for vertical direction
- scrollbar seems to not be reversed

🤔

reyamir:
You can check my `reverse_scroll` example, when use reverse mode you need to set `ScrollPosition::Start`
https://github.com/user-attachments/assets/a7a9381a-cb23-4027-928d-46f1e6521e05

this is same behavior with discord or other chat apps

https://github.com/user-attachments/assets/79e6d54b-22ea-4335-8903-6492439cc902

I see no use cases for `horizontal direction`, but jkwe can add it by simple change here:
```rust
let align = match reverse && direction_is_vertical {
  true => "end",
  false => "start",
};
```

PR #797, opened 2024-07-16
enhancement: add support for `fill` in `svg`

### LINKED ISSUE #796: enhancement: Support for `fill` attribute in `svg`

Aiving:
Closes #796


marc2332 (changes requested):
Why not `fill` instead of `color`? The former is used for text

Aiving:
> Why not `fill` instead of `color`? The former is used for text

I thought that adding a new property to `StyleState`, `AttributeName`/`Type` etc. would be a bit overkill, but if there is no problem with that I can implement it right now

Aiving:
@marc2332 done!

crates/core/src/elements/svg.rs
```diff
@@ -27,10 +30,40 @@ impl ElementUtils for SvgElement {
...
+                        }
+                        Fill::LinearGradient(gradient) => {
+                            paint.set_shader(gradient.into_shader(area));
+                        }
+                    }
+
+                    canvas.draw_paint(&paint);
+                }
+
                 canvas.restore();
```
geom3trik, line 67: You've got to restore here but no corresponding save because it's been removed

crates/elements/src/definitions.rs
```diff
@@ -479,6 +479,8 @@ builder_constructors! {
     /// }
     /// ```
     svg {
+        #[doc = include_str!("_docs/attributes/color.md")]
```
marc2332, line 482: You will need to create the doc file as well
```suggestion
        #[doc = include_str!("_docs/attributes/fill.md")]
```

marc2332 (changes requested):
Can you also add an example?

crates/elements/src/_docs/attributes/fill.md
```diff
@@ -0,0 +1,21 @@
+The `fill` attribute allows you to specify the fill color for the `svg`.
+
+You can learn about the syntax of this attribute in [`Color Syntax`](crate::_docs::color_syntax).
+
+### Example
+
+```rust, no_run
+# use freya::prelude::*;
+fn app() -> Element {
+    let svg_content = include_str!("./ferris.svg");
```
marc2332, line 10:
```suggestion
    let svg_content = include_str!("../../../examples/settings.svg");
```

marc2332:
Does this implementation really make sense?

For example, given this code:

```rs
<svg width="525" height="190" viewBox="0 0 525 190" fill="none" xmlns="http://www.w3.org/2000/svg">
    <rect width="250" height="190" fill="#D9D9D9"/>
    <rect x="275.5" y="0.5" width="249" height="189" stroke="black"/>
</svg>
```

Both `rect` will have their colors (`fill="#D9D9D9"` and `stroke="black"`) overrode when using this new `fill` attribute

Do you think that makes sense?

marc2332 (changes requested):
btw

Aiving:
> Does this implementation really make sense?
> 
> For example, given this code:
> 
> ```rust
> <svg width="525" height="190" viewBox="0 0 525 190" fill="none" xmlns="http://www.w3.org/2000/svg">
>     <rect width="250" height="190" fill="#D9D9D9"/>
>     <rect x="275.5" y="0.5" width="249" height="189" stroke="black"/>
> </svg>
> ```
> 
> Both `rect` will have their colors (`fill="#D9D9D9"` and `stroke="black"`) overrode when using this new `fill` attribute
> 
> Do you think that makes sense?

Honestly, I don't think so. Based on the Skia source code, there is a function [`resolveSvgColor`](https://github.com/google/skia/blob/938b8b8efdface485b737512805582445015fb14/modules/svg/src/SkSVGRenderContext.cpp#L471) where some ``fPresentationContext`` is used, where you can set the color used for `currentColor`. However, it is in `SkSVGRenderContext` which I can't find in `skia-safe`. In case I do, I will try to make a PR in `skia-safe` that adds the `SkSVGRenderContext`.

crates/state/src/style.rs
```diff
@@ -36,6 +37,7 @@ use crate::{
 #[derive(Default, Debug, Clone, PartialEq, Component)]
 pub struct StyleState {
     pub background: Fill,
+    pub fill_stroke: (Option<Color>, Option<Color>),
```
marc2332, line 40: Why is it a tuple? better to just put each color in its own field right? Don't see the need for a tuple
```suggestion
    pub svg_fill: Option<Color>,
    pub svg_stroke: Option<Color>,
```

crates/state/src/style.rs
```diff
@@ -58,6 +60,22 @@ impl ParseAttribute for StyleState {
                     self.background = Fill::parse(value)?;
                 }
             }
+            AttributeName::Fill => {
+                if let Some(value) = attr.value.as_text() {
+                    if value == "none" {
+                        return Ok(());
+                    }
+                    self.fill_stroke.0 = Some(Color::parse(value)?);
```
marc2332, line 68:
```suggestion
                    self.svg_stroke = Some(Color::parse(value)?);
```

crates/state/src/style.rs
```diff
@@ -58,6 +60,22 @@ impl ParseAttribute for StyleState {
...
+                }
+            }
+            AttributeName::Stroke => {
+                if let Some(value) = attr.value.as_text() {
+                    if value == "none" {
+                        return Ok(());
+                    }
+                    self.fill_stroke.1 = Some(Color::parse(value)?);
```
marc2332, line 76:
```suggestion
                    self.svg_fill = Some(Color::parse(value)?);
```

crates/state/src/style.rs
```diff
@@ -132,6 +150,7 @@ impl State<CustomAttributeValues> for StyleState {
     const NODE_MASK: NodeMaskBuilder<'static> =
         NodeMaskBuilder::new().with_attrs(AttributeMaskBuilder::Some(&[
             AttributeName::Background,
+            AttributeName::Fill,
```
marc2332, line 153: You are missing Stroke right?

```suggestion
            AttributeName::Fill,
            AttributeName::Stroke,
```

crates/core/src/elements/svg.rs
```diff
@@ -34,6 +38,13 @@ impl ElementUtils for SvgElement {
                 let mut root = svg_dom.root();
                 root.set_width(svg::Length::new(100.0, svg::LengthUnit::Percentage));
                 root.set_height(svg::Length::new(100.0, svg::LengthUnit::Percentage));
+                root.set_color(font_style.color);
+                if let Some(color) = node_style.fill_stroke.0.as_ref() {
```
marc2332, line 42:
```suggestion
                if let Some(color) = node_style.svg_stroke.as_ref() {
```

crates/core/src/elements/svg.rs
```diff
@@ -34,6 +38,13 @@ impl ElementUtils for SvgElement {
...
+                    root.set_fill(svg::Paint::from_color(*color));
+                }
+                if let Some(color) = node_style.fill_stroke.1.as_ref() {
```
marc2332, line 45:
```suggestion
                if let Some(color) = node_style.svg_fill.as_ref() {
```

marc2332 (changes requested):

Aiving:
@marc2332 done!

PR #811, opened 2024-08-10
feat: Incremental rendering

marc2332:
Closes https://github.com/marc2332/freya/issues/247

This PR adds incremental rendering support for Freya. Allowing to render only what has changed + what has been affected by the changes. This is a major improvement for rendering performance as Freya used to do full re-renders of the whole tree in every render frame, thus slowing down as the tree of nodes increased, even if only 1 small node changed.

## To Do
- [x] Layout-based dirty area invalidation
- [x] Rotation-based dirty area invalidation
- [x] Border-based dirty area invalidation
- [x] Shadows-based (text included) dirty area invalidation
- [x] Incremental opacities and rotations inheritance
- [x] Enable parallel processing of style states in Nodes
- [x] Improved logging
- [x] Refactored the rendering pipeline
- [x] Research potential optimizations:
	- ~~[ ] Consider viewports to reduce the dirty area~~
	- ~~[ ] Use `dashmap` for `hashmaps/hashsets` currently wrapped in `Arc<Mutex<T>>`.~~
	- [x] Ignore non-drawable Nodes when checking if they affect the dirty area
- [x] More unit tests
- [x] Clean up code

## Findings
- https://github.com/marc2332/freya/issues/815
- https://github.com/marc2332/freya/issues/814


marc2332:
better idea:

have a secondary canvas in which to render all the mutated layers with a clip of the affected area, and then just write the secondary canvas to the main

kinda inspired in vizia

marc2332:
to do: test svgs that have a fixed size

PR #813, opened 2024-08-10
Add opengl_rtt example.

hellozyemlya:
- wraps parameters of CanvasRunner to struct, just in case more rendering information will be passed in future
- adds simple opengl_rtt example:
  - messing with raw opengl to render triangle
  - renders to texture
  - using raw opengl texture to draw on skia canvas


marc2332 (approved):
The code looks nice! I'll try it asap. Nice work! and sorry for the delay

PR #836, opened 2024-08-31
feat: per-side border widths

Tropix126:
Adds support for individual border-widths per side. Made based on observation of figma's stroke rendering system.

## Basic Rundown

Rather than using Skia's stroking API, we instead draw the border as a second path (the "border path") with either a larger or smaller corner radius (the offset for roundness is calculated by `RectElement::border_corner_radius_offset`). We then mask out a smaller rect (the "border clip path") from the inside of the border rect.

What this gives us is the ability to offset the border rect allowing for different widths on each side.

## Current Limitations
- ~~The parser requires the user to input exactly four width values and no less currently. This means all examples and devtools are currently broken. I need to find a way to parse the widths in a clean way.~~
- ~~Devtools are broken as mentioned before.~~
- ~~Zero-width borders bleed through from under the rect if they happen to render on a subpixel.~~ This is a [skia bug](https://neugierig.org/software/chromium/notes/2010/07/clipping.html) and will be ~~fixed~~ avoided once pixel snapping is implemented. (#837)

## Some cursed showcases

Superellipse roundness with variable widths on each side. (this rendering is actually more correct than figma's, which doesn't do squircle calculations on borders :D)
image

Variable widths with variable corner radius.
image


Tropix126:
Linux build is failing for reasons I can't determine. The build also took 49 minutes :p

marc2332:
> Linux build is failing for reasons I can't determine. The build also took 49 minutes :p

Check the skia bindings

marc2332 (changes requested):
Would be nice to have the border.rs example update showing this new thing right?

marc2332 (changes requested):
this branch:
image

main:
image

🤔

Tropix126:
Rendering quality has been fixed. Example has been updated as part of #889 (which is dependent on this PR).

crates/state/src/values/border.rs
```diff
@@ -16,18 +17,60 @@ pub enum BorderStyle {
...
 impl Border {
     #[inline]
     pub fn is_visible(&self) -> bool {
-        self.width > 0. && self.fill != Fill::Color(Color::TRANSPARENT)
+        !(self.width.top == 0.0
+            && self.width.left == 0.0
+            && self.width.bottom == 0.0
+            && self.width.right == 0.0)
+            && self.fill != Fill::Color(Color::TRANSPARENT)
+            && self.style != BorderStyle::None
```
marc2332, line 46:
```suggestion
        (self.width.top != 0.0
            || self.width.left != 0.0
            || self.width.bottom != 0.0
            || self.width.right != 0.0)
            && self.fill != Fill::Color(Color::TRANSPARENT)
            && self.style != BorderStyle::None
```

marc2332 (changes requested):

examples/border.rs
marc2332, line 1: Could you update this example to contain a multi side widths border?

marc2332: rest lgtm, will merge after this

marc2332 (changes requested):

marc2332 (approved):

PR #837, opened 2024-09-01
feat: round to physical pixel boundaries while rendering

Tropix126:
This is a rewrite of #245 that's up to date. Same deal - it just rounds all elements to render on physical pixel boundaries rather than relying subpixel antialiasing which results in blurry borders or incorrect rendering in the case of #836.

### DIFF
--- a/crates/core/src/elements/image.rs
+++ b/crates/core/src/elements/image.rs
@@ -21,7 +21,7 @@ impl ElementUtils for ImageElement {
         _default_fonts: &[String],
         _scale_factor: f32,
     ) {
-        let area = layout_node.visible_area();
+        let area = layout_node.visible_area().round();
         let node_style = node_ref.get::<StyleState>().unwrap();
         let node_references = node_ref.get::<ReferencesState>().unwrap();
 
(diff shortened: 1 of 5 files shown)


marc2332 (changes requested):
The switch animation looks weird now, and I don't think disabling the rounding just for the switch is a good solution either.

Tropix126:
> The switch animation looks weird now, and I don't think disabling the rounding just for the switch is a good solution either.

Any suggestions on what I should do here instead? I'll look into the switch animation.

Tropix126:
I've fixed the alignment of the switch by disabling rounding. I'm not sure what's causing the animation to look weird (it seems fine to me although I might be missing something). I'm open to suggestions for alternatives other than disabling rounding on the switch, but any solution other than disabling it is probably going to inevitably mess up on some scale factors.

marc2332:
Try resizing the window vertically.

Eventually you will see how the inner circle gets missaligned

image


When this happens and you run the animation, it will appear as:

https://github.com/user-attachments/assets/6146c77e-616e-4ee0-9b8b-0c380f03543e

Tropix126:
Hmm... Okay try now, I think it should be fixed on switch.

marc2332:
Closing this to inactivity and to be honest not being IMO good enough solution at the moment. This may be revisited in the future

Issue #11, opened 2022-09-26
tracking: calc() attribute


marc2332:
This issue tracks the state of the `calc()` function.

## Summary

`calc()` is inspired by the CSS's [spec](https://developer.mozilla.org/en-US/docs/Web/CSS/calc). It allows specifying a certain calculation instead of using a static value. 

For example, this way we can draw a rectangle whose width will be the half of the window plus 100:

```rust
fn app(cx: Scope) -> Element {
    cx.render(rsx!(rect {
        background: "black",
        width: "calc(50% + 100)",
        height: "100%",
    }))
}
```

## Downsides
- Calculations are made in order, so there is no priority for divisions or multiplications. 
For example, the expected result of `calc(100 + 250 * 2)` would be 600, but it's 700 since it goes one by one.

PR #838, opened 2024-09-01
feat: Calc operator precedence

maiadere:
This PR adds operator precedence which hasn't been implemented in #11. It also makes it easier to add parentheses support in the future (after #785 is merged). `calc(100 + 250 * 2)` will now return 600.


crates/torin/src/values/size.rs
marc2332, line 1: Hey! Cool work. Do you mind assing some comments explaining the code?

crates/torin/src/values/size.rs
```diff
@@ -193,50 +196,79 @@ impl std::fmt::Display for DynamicCalculation {
...
+            current: None,
+        }
+    }
+
+    pub fn evaluate(&mut self) -> Option<f32> {
+        self.current = self.calcs.next();
+        self.expr(0)
+    }
+
+    pub fn expr(&mut self, precedence: usize) -> Option<f32> {
```
marc2332, line 219: Also, maybe update the function names, "expr" is quite vague

crates/torin/src/values/size.rs
```diff
@@ -193,50 +196,79 @@ impl std::fmt::Display for DynamicCalculation {
...
+        if self.current.is_some() {
+            return None;
+        }
 
-                calc_with_op(val, prev_op);
+        Some(lhs)
+    }
 
-                prev_op = None;
+    pub fn term(&mut self) -> Option<f32> {
```
marc2332, line 247: Same for this, "term" doesn't really explain much

marc2332 (changes requested):

crates/torin/src/values/size.rs
noeko: Would calling it `value` be better? Since currently it's either percentage or pixels. I have seen it being called atom in a lot of grammars and parsers, but I think value would be more intuitive? And would renaming function names to `parse_value` and `parse_expression` be fine?

crates/torin/src/values/size.rs
noeko: I added some

marc2332 (approved):
Thanks :) 🦀

PR #840, opened 2024-09-02
Update animated_sidebar.rs

seppelin:
The node height is at the beginning of the animation a few frames 0. This leads to flickering when moving up because the offset is 0 an not the page height. This is an easy fix but maybe it would be better to fix the node_size because this is very unintuitive. This problem exists in the tabs as well I believe. It's easily visible when using ease In.

### DIFF
--- a/examples/animated_sidebar.rs
+++ b/examples/animated_sidebar.rs
@@ -73,7 +73,7 @@ fn FromRouteToCurrent(from: Element, upwards: bool) -> Element {
                 width: "fill",
                 Expand { {from} }
             }
-        )
+        );
     }
 
     let offset = height - (offset * height);
--- a/examples/animated_sidebar.rs
+++ b/examples/animated_sidebar.rs
@@ -62,6 +62,20 @@ fn FromRouteToCurrent(from: Element, upwards: bool) -> Element {
     let offset = animations.get().read().as_f32();
     let height = node_size.area.height();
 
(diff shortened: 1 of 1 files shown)


marc2332:
I don't see any issue in my end, do you mind sending a video or something?

seppelin:
https://github.com/user-attachments/assets/dc80d847-1ebe-4361-a644-a620d84ade2b

marc2332:
I think the issue might be somewhere else.

marc2332:
Hey @seppelin ! I believe I just pushed a better fix -> https://github.com/marc2332/freya/commit/7b84d20fff606d08b5ac08d093821bf3a961bdae

Can you try now with the main branch? 😄

seppelin:
Yeah, it works perfectly and it's a nice solution! Thank you

seppelin:
Do you know why the size of the inner node is 0 for, at least at my pc, 3 frames?

marc2332:
> Do you know why the size of the inner node is 0 for, at least at my pc, 3 frames?

use_node works asynchronously so it might not have the value filled in until the layout is processed. Also, your component rerunning doesn't equal a frame, your component may in fact run multiple times and no frame might actually be renderered.

Should I close this then?

seppelin:
Yes

marc2332:
> Yes

Thanks for bringing this to my attention btw! I always appreciate any kind of feedback or help.

😄

PR #841, opened 2024-09-02
refactor: overhauled accessibility system via hook

Tropix126:
Overhauls freya's accessibility APIs.
- Adds `use_accessibility`, which replaces and extends the previous `use_focus` hook. This gives the user access to most AccessKit `NodeBuilder` properties and flags.
- Removes existing accessibility attributes in favor of the hook.
- Attempts to derive as much data as possible from the node's styles to fill in accesskit properties.

TODO:
- [x] Implement an `ActionHandler` in Freya for handling commands from assistive technology.
- [ ] Expose a way for users to add actions to their accessibility nodes.
- ~~[ ] Figure out inline text spans/measurement and how they play into this.~~ Will be addressed later, need to hook skia text measurement APIs for that :/.
- [ ] Determine if AccessKit automatically infers properties such as `position_in_set` if unset. I know browsers do this, but it might be a different story here and we might have to do it ourselves (which would suck).
- [ ] Update components. This actually a HUGE undertaking if we want to do it right, and im leaning towards stripping out all accessibility from `freya_components` addressing this in future PRs, since the current state of accessibility in components is very incomplete or incorrect anyways.
- [ ] Update tests.
- [ ] Update examples.


Tropix126:
Closing. This will be reimplemented in smaller PRs.

PR #855, opened 2024-09-08
feat: infer accesskit properties from node state

Tropix126:
This PR fills out the following AccessKit node properties based on known node state:
- `transform` (rotation)
- `clips_children` (overflow)
- `foreground_color`, `background_color`
- `is_line_breaking_object` (has block-level layout)
- `font_size`
- `font_family` (only for nodes changing the family, child nodes already inherit)
- `bold`
- `text_align`
- `text_direction` (always LTR currently)
- `italic` (for `FontSlant` `Italic` and Oblique)
- `text_decoration`
- `underline`
- `overline`
- `strikethrough`

It also adjusts the way that the `Focus` action is applied for focusable nodes. This is the previous code:
```rs
// Set focusable action
if node_accessibility.focusable {
    builder.add_action(Action::Focus);
} else {
    builder.add_action(Action::Default);
    builder.set_default_action_verb(DefaultActionVerb::Focus);
}
```
Previously, if the node wasn't focusable, a screenreader would offer the user a "default action" (which could mean anything, but typically means "click") whether or not the node was clickable or not. It also set the action verb for the default action to be announced as "focus", which isn't correct since the action does effectively nothing (freya only handles `Action::Focus` in its event handling code).

Instead, i've removed the `else` clause which means that non-focusable nodes will have no actions by default. In the future we will want to allow users the ability to add actions to their elements, but this is out of scope of this PR.

### DIFF
--- a/crates/core/src/accessibility/tree.rs
+++ b/crates/core/src/accessibility/tree.rs
@@ -18,10 +18,10 @@ use accesskit::{
 use freya_common::AccessibilityDirtyNodes;
 use freya_engine::prelude::{
     Color,
+    Slant,
     TextAlign,
     TextDecoration,
     TextDecorationStyle,
-    Slant,
 };
 use freya_native_core::{
     node::NodeType,
--- a/crates/core/src/accessibility/tree.rs
+++ b/crates/core/src/accessibility/tree.rs
@@ -4,23 +4,16 @@ use std::sync::{
 };
 
 use accesskit::{
-    Action,
-    DefaultActionVerb,
-    Node,
-    NodeBuilder,
-    NodeId as AccessibilityId,
-    Rect,
-    Role,
-    Tree,
-    TreeUpdate,
+    Action, Affine, Node, NodeBuilder, NodeId as AccessibilityId, Rect, Role, TextDirection, Tree, TreeUpdate
 };
 use freya_common::AccessibilityDirtyNodes;
 use freya_native_core::{
-    prelude::NodeImmutable,
-    tags::TagName,
-    NodeId,
+    node::NodeType, prelude::NodeImmutable, tags::TagName, NodeId
 };
-use freya_node_state::AccessibilityNodeState;
+use freya_node_state::{
+    AccessibilityNodeState, Fill, FontStyleState, OverflowMode, StyleState, TransformState
+};
+use freya_engine::prelude::{Color, TextAlign, TextDecoration, TextDecorationStyle};
 use rustc_hash::{
     FxHashMap,
     FxHashSet,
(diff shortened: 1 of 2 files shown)


crates/core/src/accessibility/tree.rs
marc2332, line 1: Just general question, does it make sense to always set as many things as a node has? e.g foreground, font size, font family, etc?

crates/core/src/accessibility/tree.rs
```diff
@@ -280,12 +295,133 @@ impl AccessibilityTree {
...
+            TextAlign::Right | TextAlign::End => accesskit::TextAlign::Right,
+        });
+
+        // TODO: Adjust this once text direction support other than RTL is properly added
+        builder.set_text_direction(TextDirection::LeftToRight);
+
+        // Set italic property for italic/oblique font slants
+        match font_style_state.font_slant as u32 {
+            // Italic | Oblique
+            1 | 2 => builder.set_italic(),
```
marc2332, line 392: why are you using numbers instead of enum variants?

crates/core/src/accessibility/tree.rs
```diff
@@ -280,12 +295,133 @@ impl AccessibilityTree {
...
+            {
+                builder.set_font_family(font_style_state.font_family.join(", "));
+            }
+        } else {
+            // Element has no parent elements, so we set the initial font style.
+            builder.set_font_family(font_style_state.font_family.join(", "));
+        }
+
+        // Set bold flag for weights above 700
+        if font_style_state.font_weight > 700.into() {
```
marc2332, line 373: You will need to update `mocked.rs` so this doesn't make the CI error

marc2332 (changes requested):

crates/core/src/accessibility/tree.rs
Tropix126: Yeah, browsers do it and it could be potentially useful for assistive tech. Screenreaders might not necessarily use it (although they might choose to emphasize based on font style), but other things such as magnifiers or utilities for colorblind people may use these properties depending on the platform. I think its good to fill them out if we can.

crates/core/src/accessibility/tree.rs
Tropix126: Sorry, some of this code was brought over from #841 (which is now dead). I think at the time I couldn't get `Slant` in scope. I'll get that changed.

crates/core/src/accessibility/tree.rs
marc2332: Nice

marc2332 (approved):

marc2332:
🔥

PR #873, opened 2024-09-11
fix(torin): Ignore absolute nodes for spacing and alignments

marc2332:
Completely ignore absolute nodes when measuring the required spacing between children and also for axis alignments

### DIFF
--- a/crates/torin/src/measure.rs
+++ b/crates/torin/src/measure.rs
@@ -345,8 +345,7 @@ where
                                 None
                             }
                         },
-                    )
-                    .unwrap();
+                    );
 
                 let inner_area = initial_phase_inner_area;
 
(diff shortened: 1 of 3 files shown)


marc2332:
Holy 💩 the benchmark 💀

crates/torin/src/measure.rs
```diff
@@ -438,18 +472,18 @@ where
...
-                parent_node,
-                area,
-                inner_area,
-                inner_sizes,
-                &child_areas.area,
-                &child_data,
-                children_len,
-                child_n,
-            );
+            // Stack this s child into the parent
```
marc2332, line 475:
```suggestion
            // Stack this child into the parent
```

PR #889, opened 2024-09-15
feat: multiple borders

Tropix126:
Depends on #836.

- Removes the `border_align` attribute.
	- This was necessary to do since individual borders can have their own alignments. Also, since the border style argument can only be `solid` or `none` anyways, this seemed reasonable.
	- If more styles other than solid (dashed, dotted, etc...) are eventually added (adding these at this point would be pretty ridiculously complex) they can be added as an optional argument e.g. `1 inner dotted red`.
	- Replaces the redundant `<style>` field with alignment intead. Syntax is now `<width> <width?> <width?> <width?> <align> <fill>` e.g. `1 inner red`.
- Adds support for styling multiple borders on a single `rect` (similar to how `shadow` already works).

```rs
rect {
    width: "80",
    height: "80",
    corner_radius: "2",
    background: "rgb(0, 0, 0)",
    border: "6 outer red, 5 outer orange, 4 outer yellow, 3 outer green, 2 outer blue, 1 outer purple",
    margin: "8"
}
```

image


crates/elements/src/_docs/attributes/border.md
marc2332, line 1: This file still mentions the old syntax and `border_align`

marc2332 (changes requested):

marc2332 (approved):
Lovely.

PR #902, opened 2024-09-20
Redo animations

RobertasJ:
This should make it a lot easier to make animations that consist of multiple easing function run one after another.

### DIFF
--- a/crates/hooks/src/lib.rs
+++ b/crates/hooks/src/lib.rs
@@ -8,6 +8,7 @@ mod text_editor;
 mod theming;
 
 mod use_activable_route;
+pub mod use_animate;
 mod use_animation;
 mod use_asset_cacher;
 mod use_canvas;
--- a/crates/hooks/src/use_animate.rs
+++ b/crates/hooks/src/use_animate.rs
@@ -1,13 +1,36 @@
-use std::{fmt::Debug, time::Instant};
+use std::{
+    fmt::Debug,
+    time::Instant,
+};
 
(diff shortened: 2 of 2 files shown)


RobertasJ:
It already is a replacement for AnimValue, we could just replace AnimValue with it and make an AnimVolor variant for that. 

I also have no idea how to fix the is_running method on it rn as idk how the index is given.

marc2332:
> It already is a replacement for AnimValue, we could just replace AnimValue with it and make an AnimVolor variant for that.
> 
> I also have no idea how to fix the is_running method on it rn as idk how the index is given.

What do you mean by a variant? Like an enum variant? Wouldn't that be the same as having two structs?

RobertasJ:
I meant it as an another struct, its just simmilar enough that i called it a variant of the original xd

marc2332:
> I meant it as an another struct, its just simmilar enough that i called it a variant of the original

yeah but the thing is that the only difference is that you can have multiple steps, I just feel like we could just move this into the other two structs somehow

RobertasJ:
Yeah, we can do that

marc2332:
I feel like would be better as a crate, if there is anything that it needs that is not yet exported we can make it public in this very own PR, wdyt

RobertasJ:
This sounds pretty good, but i do wonder if i could redo the animation logic so that it reverses the animation correctly instead of doing a jump

RobertasJ:
I think its best for me to close this as its a whole different hook, its better as its own crate

PR #921, opened 2024-09-25
fix: linear gradient angles

### LINKED ISSUE #913: bug: Angle of linear gradient does not rotate correctly
Freya ( `background: "linear-gradient(90deg, rgb(255,255,255) 0%, rgb(0,0,0) 100%)",`):
image

btw, changing the angle from 90 to 45 does not fix the issue.

image

Browsers
image

Tropix126:
Closes #913


marc2332:
So nice! Thanks @Tropix126 !

Issue #57, opened 2022-11-24
enhancement: Zoom  support


marc2332:
Maybe add support for Zoom via an attribute?

PR #931, opened 2024-09-27
feat: Custom scale factor shortcuts

marc2332:
Closes https://github.com/marc2332/freya/issues/57


marc2332:
TODO: Use command in mac

Issue #966, opened 2024-10-06
bug: Accessibility adapter doesn't get focus on window creation


DataTriny:
When navigating with the keyboard, the focus seem to not be reported to assistive technologies, hence nothing is announced if using a screen reader for instance. I can see visually that UI elements get the focus though.

Tested examples:

- accessibility,
- keyboard_navigation

Testing environment: Windows 11 and NVDA (screen reader)

DataTriny:
Interestingly, if I move the system focus away from the Freya app window and then focus it back, keyboard navigation seem to work as expected.

marc2332:
I actually also got this a few times and just thought it was normal. I guess it is not, will have a look!

marc2332:
Weird, the tree update is definitely going through in `update_if_active`, yet, the screen reader doesn't focus it

marc2332:
By default, the window adapter seems to be initialized internally with `State::Inactive`, could be related?

marc2332:
Also, `SubclassImpl` passes a false to the window focus when initialized 

https://github.com/AccessKit/accesskit/blob/fc622fe7657c80a4eedad6f6cded11d2538b54d5/platforms/windows/src/subclass.rs#L82

DataTriny:
Very quickly looking through the code, I'm not sure you open the winit window hidden, then initialize AccessKit, then show the window. This is documented [here](https://docs.rs/accesskit_winit/latest/accesskit_winit/struct.Adapter.html#method.with_event_loop_proxy).

marc2332:
ooooops! That was it.

marc2332:
Just opened https://github.com/marc2332/freya/pull/967 @DataTriny

PR #967, opened 2024-10-06
fix: Create accesskit adapter before making window visible

marc2332:
Closes https://github.com/marc2332/freya/issues/966


DataTriny:
This appears to fix the issue.

DataTriny (approved):

marc2332:
> This appears to fix the issue.

Cool! Thanks for the report and the testing 🦀

Issue #389, opened 2023-11-19
enhancement: Ability to disable `Button`


marc2332:
- Not able to click on them
- Grayed colors

PR #968, opened 2024-10-07
enhancement: added enabled prop to button

ghost:
First step to finish #389.


crates/components/src/button.rs
```diff
@@ -255,13 +278,16 @@ pub fn ButtonBase(
         format!("1 inner {border_fill}")
     };
 
+    let opacity = if enabled { "1" } else { "0.5" };
```
marc2332, line 281: Opacity is not the way to go, it should be changed at the theme level

marc2332: Otherwise the content below the buttons will be visible

marc2332 (changes requested):

PR #986, opened 2024-10-16
fix: absolute positioning doesn't ignore the padding of the parent element

Aiving:


marc2332 (changes requested):
What about the root node when measuring it in torin.rs?

Aiving:
> What about the root node when measuring it in torin.rs?

I'm not sure I understand correctly what the question is about. Do you mean absolute positioning of the root element itself or something else?

marc2332:
> > What about the root node when measuring it in torin.rs?
> 
> I'm not sure I understand correctly what the question is about. Do you mean absolute positioning of the root element itself or something else?

Torin starts measuring from the root node by calling `measure_node`, you changed how this method is called when measuring descendants but not when is called upon the root node of the measurement.

By root node I don't specifically refer to the root node of the app but to the root node from where the measuring starts, which could be different is the measure is incremental.

marc2332:
Perhaps this your way is not a good approach, if the 3 places where this function is called require some special logic maybe its better to move it inside or create a simple function to abstract it

marc2332:
I don't agree with this, if an element is absolute I still think it should consider the parent's padding. I recently added the `global` position too, maybe that fits your use case better?

marc2332:
closing to inactivity, feel free to open it again in the future again

PR #988, opened 2024-10-20
Add prefixes and parentheses to calc function

RobertasJ:
Added support for:
- parenthesses syntax
- implicit multiplication:
  - calc(1.5(5 + 7)
- prefixes:
  - calc(-(5 - 10))


crates/torin/src/values/size.rs
```diff
@@ -267,35 +271,88 @@ impl<'a> DynamicCalculationEvaluator<'a> {
...
+            } else {
+                return None;
+            }
+            last_is_seperator = seperator;
+        }
+        if let Some(prefix) = prefix {
+            match prefix {
+                DynamicCalculation::Add => lhs,
+                DynamicCalculation::Sub => lhs.map(|v| v * -1.0),
+                _ => unreachable!(),
```
marc2332, line 295: Maybe put some message in the `unreachable!()`, although it would be better if this was integrated into the parsing error reporting instead of panicking, but this is better done in another PR

marc2332 (changes requested):

crates/torin/src/values/size.rs
RobertasJ: Added something there now

crates/torin/src/values/size.rs
```diff
@@ -275,35 +279,88 @@ impl<'a> DynamicCalculationEvaluator<'a> {
...
+            } else {
+                return None;
+            }
+            last_is_seperator = seperator;
+        }
+        if let Some(prefix) = prefix {
+            match prefix {
+                DynamicCalculation::Add => lhs,
+                DynamicCalculation::Sub => lhs.map(|v| v * -1.0),
+                _ => unreachable!("make sure to add the prefix here"),
```
marc2332, line 303: Not sure what you mean by this message, can this actually happen? if this can happen we should not be panicking

crates/torin/src/values/size.rs
```diff
@@ -275,35 +279,88 @@ impl<'a> DynamicCalculationEvaluator<'a> {
...
 
-    /// Parse and evaluate the value with the following grammar:
+    /// Parse and evaluate the term, implements implicit multiplication. only parenthesis count as
+    /// a seperator, so syntax like 50 50 isnt correct, but 50(50) is because the parenthesis act
+    /// as a seperator
+    fn parse_term(&mut self) -> Option<f32> {
+        let prefix = self.parse_prefix()?;
+        let mut lhs = None;
+        // set to true so that the first value is multiplied and counts as normal syntax
+        let mut last_is_seperator = true;
```
marc2332, line 289:
```suggestion
        let mut last_is_separator = true;
```

crates/torin/src/values/size.rs
RobertasJ: it cannot happen, if this happens it means the person adding a new prefix would have forgotten to handle it here

marc2332 (changes requested):

marc2332 (changes requested):
for a second I thought I was going crazy but I think you forgot to handle parsing right? `parse_calc`

crates/torin/src/values/size.rs
RobertasJ: fixed

PR #1006, opened 2024-11-24
Fix scale factor behavior in `calc()`

RobertasJ:


crates/torin/src/values/size.rs
```diff
@@ -166,7 +166,10 @@ impl Scaled for Size {
         match self {
             Size::Pixels(s) => *s *= scale_factor,
             Size::DynamicCalculations(calcs) => {
-                calcs.iter_mut().for_each(|calc| calc.scale(scale_factor));
+                calcs.insert(0, DynamicCalculation::OpenParenthesis);
+                calcs.push(DynamicCalculation::ClosedParenthesis);
+                // implicit multiplication
```
marc2332, line 171: Lets make it no implicit for the sake of not under-simplifying

marc2332 (changes requested):

crates/torin/src/values/size.rs
RobertasJ: Done :D

marc2332 (changes requested):
Can you add some units asserts that check that scaling a size using calc() works just fine with different sizes of scale factors? Also format the code :D

marc2332 (approved):

marc2332:
Nice work!

PR #1007, opened 2024-11-30
calc easier to extend syntax

RobertasJ:
A pr where instead of using symbols for custom values, we have variables instead, list of variables:
- root
- root.cross
- root.width
- root.height
- parent
- parent.cross
- parent.width
- parent.height
- scale
in addition to that there are now new functions one can use within calc:
- min(value1, value2, value3, ...)
- max(value1, value2, value3, ...)
- clamp(minimum_value, value, maximum_value)

an always square rect element example:
```rust
rsx! {
    rect {
        width: "100", // imagine this value changing arbitrarily
        height: "400", // imagine this value changing arbitrarily
        rect {
            width: "calc(min(parent, parent.cross))", 
            height: "calc(min(parent, parent.cross))"
        }
    }
}
```
i marked it as a draft as im sure this is pretty new. might be interesting to discuss if this syntax would be good or not.


RobertasJ:
@marc2332 wondering on how you would want me to change this in order to limit the possibility of invalid states.

RobertasJ:
i will close this pr since i dont think its quite needed atm.

PR #1013, opened 2024-12-17
feat: New `AnimatedPosition` component

marc2332:
Depends on https://github.com/marc2332/freya/pull/1015


marc2332:
Notes: I don't like the name of `fixed`, maybe `global` is better?

Issue #1029, opened 2024-12-29
enhancement: Support auto-sized centered text


marc2332:
Support text that has an auto size and has a centered alignment

Currently it expands to 100% of the available width if you use a `text_align: "center"`

PR #1030, opened 2024-12-29
feat: Support auto size for text with non-start alignment

marc2332:
Closes https://github.com/marc2332/freya/issues/1029


marc2332:
todo: tests!

PR #1032, opened 2025-01-01
feat: Simplify `use_asset_cacher` hook

marc2332:


crates/hooks/Cargo.toml
```diff
@@ -41,6 +41,7 @@ nokhwa = { version = "0.10.7", features = ["input-native"], optional = true }
 paste = "1.0.14"
 bitflags = "2.4.1"
 bytes = "1.5.0"
+tracing.workspace = true
```
marc2332, line 44:
```suggestion
tracing = { workspace = true }
```

examples/app_dog.rs
marc2332, line 1: Revert this too

PR #1043, opened 2025-01-12
chore: Rust 1.83.0

marc2332:


marc2332:
Might as well just jump to 1.84.0

PR #1055, opened 2025-01-16
feat:  `no_std` support for `torin`

marc2332:


marc2332:
Too much work for something I don't think is worth atm.

PR #1066, opened 2025-01-25
feat: Improved  attributes docs

marc2332:
This allows hovering the attributes to see their docs :)


marc2332:
to do: the same but for events!

PR #1067, opened 2025-01-25
fix: Events triggering fixes

marc2332:
## Fixes
- Only update the child_node layer check on valid nodes
- Add a layer check on the nodes state 
- Check the potential events in reversed order of declaration


marc2332:
to do:
- tests
- refactor code
- stop crying
- comment the code

PR #1080, opened 2025-01-31
Add `sampling` property for `image`

Aiving:
Applications often use images and some of them can be quite large in size, but still be used in a component with a small size. At the moment when this happens, Freya simply reduces the image in the simplest way possible: through nearest sampling. Unfortunately, this sampling is only suitable for pixel art.

This PR aims to solve this problem by adding the `sampling` property to the `image` element. It will allow developer to set the sampling mode that is most pleasing to the eye (usually trilinear)

An example of how to use each of the sampling modes is also added.


crates/elements/src/attributes/image_attributes.rs
marc2332, line 1: Mind adding an explanation on how to use the sampling? Just like in `aspect_ratio`

examples/sampling.rs
marc2332, line 1: Can you rename the example file name to `image_sampling?`

marc2332 (changes requested):
Can you also add the sampling as a prop to NetworkImage?

marc2332:
rest lgtm

examples/sampling.rs
Aiving: yeah ofc

Aiving:
> Can you also add the sampling as a prop to NetworkImage?

all done!

marc2332 (approved):
🔥

PR #1097, opened 2025-02-06
feat: UI Improvements

marc2332:
- [x] More compact Button and Input components
- [x] Menu components are now navigable with keyboard and have a better design
- [x] Scrollbars are now half-collapsed unless hovered with mouse
- [x] Hover effect for table rows 
- [x] Better input handling
- [x] Disabled shadows for Button and Input components, in the built-in themes
- [x] Reduce shadow of Tooltip


marc2332:
to do: Menu component

PR #1110, opened 2025-02-18
feat: support for `current_color` in the svg attributes `fill` and `stroke`.

Aiving:


marc2332:
Love it, seems promising, will review asap

PR #1117, opened 2025-03-04
Scale values in a more usable way

RobertasJ:
This pr adds a Val type thats used to seperate the scaled and unscaled values. in the end with some logic it gets converted into an f32 by scaling the value that should be scaled and adding them to the unscaled value.

Its better to look at the code to kind of understand what it does. Have fun :D

end result is that this is the results of the calc expressions vs the old results if scaling factor is 2:
#### example 1:
`calc(2 * 5)`
result:
 - old: `4 * 10`
 - new: `(2 * 5) * 2`

#### example 2:
`calc(25% * 2)`
result:
 - old: `25% * 4`
 - new: `25% * 2`

#### example 3:
`calc(5 + 2 * 3 + 25% * 3)`
result:
 - old: `10 + 4 * 6 + 25% * 6`
 - new: `10 + (2 * 3) * 2 + 25% * 3`

#### example 4:
`calc( 2 * (25% * 2 + 10 * 3 + 2))`
result:
 - old: `4 * (25% * 4 + 20 * 6 + 4)`
 - new: `2 * (25% * 2) + (10 * 3 + 2) * 2 * 2` // the 2 * is distributed over the sums of the scaled values and the sums of the unscaled values, the scaled values then at the end get multiplied by the scale


crates/torin/src/values/size.rs
```diff
@@ -203,28 +215,283 @@ impl std::fmt::Display for DynamicCalculation {
...
+    scaling_factor: f32,
     root_value: f32,
     current: Option<&'a DynamicCalculation>,
 }
 
+#[derive(Debug, PartialEq, Clone, Copy)]
+enum Val {
+    Scaled(f32),
+    UnScaled(f32),
+    PartiallyScaled { scaled: f32, unscaled: f32 },
```
marc2332, line 227: Can you explain what each variant is for? And also explain a bit the custom SubAssign, AddAssign, etc... impls?

marc2332 (changes requested):

crates/torin/src/values/size.rs
marc2332: I mean as a comment

crates/torin/src/values/size.rs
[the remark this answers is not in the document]
RobertasJ: the values are all unscaled pixels, which get scaled at the end of the evaluate function after the final Val is returned bu parse_expression. but the % gets turned into real pixels first by taking 50% of the parent width in pixels, then wrapping that in a Val::UnScaled to mark it as pixels that shouldnt be scaled

PR #1120, opened 2025-03-17
feat: `SelectableText` component

marc2332:


marc2332:
I should probably deselect once its unfocused

Issue #1127, opened 2025-03-27
enhancement: Allow tweaking the default state of  `Accordion`


marc2332:
This way you can make an Accordion be open by default

PR #1128, opened 2025-03-27
feat: Allow passing initial state to `Accordion`

marc2332:
Closes https://github.com/marc2332/freya/issues/1127

### DIFF
--- a/crates/components/src/accordion.rs
+++ b/crates/components/src/accordion.rs
@@ -34,6 +34,9 @@ pub struct AccordionProps {
     pub children: Element,
     /// Summary element.
     pub summary: Element,
+    /// Whether its open or not initially. Default to `false`.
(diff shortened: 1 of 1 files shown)


nigeleke:
Thanks for this @marc2332 :) - however, you may want another look at the solution....
(or I've implemented something incorrectly).

It appears the "state" is open, but the AccordionBody isn't. When I click on the summary the body appears, then disappears. After that is acts as normal...

marc2332:
> Thanks for this @marc2332 :) - however, you may want another look at the solution.... (or I've implemented something incorrectly).
> 
> It appears the "state" is open, but the AccordionBody isn't. When I click on the summary the body appears, then disappears. After that is acts as normal...

Apologies, I did not think of the animation, that's where the issue is. I'll fix it asap!

PR #1152, opened 2025-04-19
feat: `use_popup` hook

marc2332:


dexterdy:
haha, that is much cleaner than what I was working on. 

I was messing around with this:

```rust
use std::thread;

use freya::prelude::*;

pub fn use_return_popup<
    T: 'static + Clone + Sync + Send,
    I: 'static + Clone + Sync + Send,
    C: Fn(I, Signal<Option<T>, SyncStorage>) -> Element,
>(
    popup: C,
) -> (Element, impl AsyncFnMut(I) -> T) {
    let mut ret_signal = use_signal_sync::<Option<T>>(|| None);
    let mut show_popup = use_signal_sync(|| false);
    let mut input = None;

    let popup = rsx! {
        if *show_popup.read() {
            {popup(input.clone().unwrap(), ret_signal)}
        }
    };

    let activate = async move |i: I| {
        input = Some(i);
        show_popup.set(true);
        let (tx, rx) = async_channel::bounded(1);
        thread::spawn(move || loop {
            let ret = ret_signal.read().clone();
            if let Some(ret) = ret {
                ret_signal.set(None);
                input = None;
                tx.send_blocking(ret);
            }
        });
        return rx.recv().await.unwrap();
    };

    (popup, activate)
}
```

It doesn't compile, lol

It is supposed to take the popup as parameter, give it a signal that it can put the return into and then, once the function was invoked, poll in a thread until the return signal was filled in.

dexterdy:
nice work though

marc2332:
> haha, that is much cleaner than what I was working on.
> 
> I was messing around with this:
> 
> ```rust
> use std::thread;
> 
> use freya::prelude::*;
> 
> pub fn use_return_popup<
>     T: 'static + Clone + Sync + Send,
>     I: 'static + Clone + Sync + Send,
>     C: Fn(I, Signal<Option<T>, SyncStorage>) -> Element,
> >(
>     popup: C,
> ) -> (Element, impl AsyncFnMut(I) -> T) {
>     let mut ret_signal = use_signal_sync::<Option<T>>(|| None);
>     let mut show_popup = use_signal_sync(|| false);
>     let mut input = None;
> 
>     let popup = rsx! {
>         if *show_popup.read() {
>             {popup(input.clone().unwrap(), ret_signal)}
>         }
>     };
> 
>     let activate = async move |i: I| {
>         input = Some(i);
>         show_popup.set(true);
>         let (tx, rx) = async_channel::bounded(1);
>         thread::spawn(move || loop {
>             let ret = ret_signal.read().clone();
>             if let Some(ret) = ret {
>                 ret_signal.set(None);
>                 input = None;
>                 tx.send_blocking(ret);
>             }
>         });
>         return rx.recv().await.unwrap();
>     };
> 
>     (popup, activate)
> }
> ```
> 
> It doesn't compile, lol
> 
> It is supposed to take the popup as parameter, give it a signal that it can put the return into and then, once the function was invoked, poll in a thread until the return signal was filled in.

Oh wow, that is indeed much complicated, using threads is an overkill for this haha

Try using this PR and feel free to leave any feedback/ideas/etc

dexterdy:
Alright, that works quite well. I can use a signal to pass information to display in the modal and then get a response through this hook. I am using an enum to distinguish between different results in the modal, which can then be used in the handler.

dexterdy:
Oh I do have 1 note:
When using this for global modals, sometimes you want to pass information from the call site to the modal. The current API doesn't have any way to support that use case. Is it maybe possible to give `open` a generic input parameter that can be used for that purpose?

crates/hooks/src/use_popup.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub fn is_open(&self) -> bool {
+        *self.open.read()
+    }
+
+    /// Read the last answer.
+    pub fn read(&self) -> ReadableRef<Signal<Option<T>>> {
+        self.value.read_unchecked()
+    }
+
+    /// Mark the popup as open an await for its response.
```
dexterdy, line 44: small typo

marc2332:
> Oh I do have 1 note: When using this for global modals, sometimes you want to pass information from the call site to the modal. The current API doesn't have any way to support that use case. Is it maybe possible to give `open` a generic input parameter that can be used for that purpose?

🫡

crates/hooks/src/use_popup.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub async fn open(&mut self) -> ReadableRef<Signal<Option<T>>> {
```
dexterdy, line 45: is there some way we can guarantee that, when the modal closes, the value is always `Some`? that way we could just return `ReadableRef<Signal<T>>`, I think.

crates/hooks/src/use_popup.rs
```diff
@@ -0,0 +1,154 @@
...
+        self.open.set(true);
+        let waker = self.waker.read().clone();
+        waker.notified().await;
+        self.open.set(false);
+        self.value.read_unchecked()
+    }
+}
+
+/// Create a popups context which can later be answered using [use_popup].
```
dexterdy, line 54: I think you meant `[use_popup_answer]` here

crates/hooks/src/use_popup.rs
marc2332: Can't do that because the value holder is created when use_popup is called and not when open() is called :( Also, you might want to close the popup without answering 💡

crates/hooks/src/use_popup.rs
marc2332: 💯

marc2332:
> Oh I do have 1 note: When using this for global modals, sometimes you want to pass information from the call site to the modal. The current API doesn't have any way to support that use case. Is it maybe possible to give `open` a generic input parameter that can be used for that purpose?

Just pushed, wdyt?

dexterdy (approved):
Nice! I like it. It works well

PR #1220, opened 2025-06-09
feat: add support for simultaneous two axis scrolling in virtualscrollview

### LINKED ISSUE #1218: enhancement: add simultaneous horizontal and vertical support to `VirtualScrollView`
It would be nice if `VirtualScrollView` supports simultaneous horizontal and vertical scrolling when the `direction` prop is not set, like the regular `ScrollView`.

kyteidev:
Closes #1218 

I was thinking of adding a "both" value for `direction` prop, so `horizontal` and `vertical` would only allow one axis scrolling, while `both` allows for two axis scrolling. In regular `ScrollView`, `direction` doesn't seem to do anything


crates/components/src/scroll_views/virtual_scroll_view.rs
```diff
@@ -428,8 +429,24 @@ pub fn VirtualScrollView<
         .map(|f| f.0 == Axis::Y)
         .unwrap_or_default();
 
-    let offset_y_min = (-corrected_scrolled_y / item_size).floor() * item_size;
-    let offset_y = -corrected_scrolled_y - offset_y_min;
+    let (offset_x, offset_y) = match direction.as_str() {
```
marc2332, line 432: I happen to have made this change also in https://github.com/marc2332/freya/pull/1219/files, I hope it is ok if I merge my pr first

crates/components/src/scroll_views/virtual_scroll_view.rs
```diff
@@ -222,27 +222,30 @@ pub fn VirtualScrollView<
     let mut focus = use_focus();
     let applied_scrollbar_theme = use_applied_theme!(&scrollbar_theme, scroll_bar);
 
-    let inner_size = item_size * length as f32;
+    let (inner_width, inner_height) = match direction.as_str() {
```
marc2332, line 225: That's how I was doing it in my branch too, lgtm

crates/components/src/loader.rs
marc2332, line 1: What's up with this changes? haha, lets revert it

crates/components/src/scroll_views/virtual_scroll_view.rs
kyteidev: sure

kyteidev:
I’ll fix the changes a bit later today, maybe in like 2 hours or so

kyteidev:
I hope it is OK that I replaced the changes in #1219 to virtual_scroll_view.rs, because they only allowed scrolling on one axis at a time.

crates/components/src/scroll_views/virtual_scroll_view.rs
```diff
@@ -430,16 +431,20 @@ pub fn VirtualScrollView<
 
     let (offset_x, offset_y) = match direction.as_str() {
         "vertical" => {
-            let min = (-corrected_scrolled_y / item_size).floor() * item_size;
-            let offset = -(-corrected_scrolled_y - min);
+            let offset_y_min = (-corrected_scrolled_y / item_size).floor() * item_size;
+            let offset_y = -corrected_scrolled_y - offset_y_min;
+
+            let offset_x = -corrected_scrolled_x;
```
marc2332, line 437: The double - here looks unnecessary, no? You also have one in the attribute in L474, maybe just remove both

kyteidev: If I understood correctly, do you want me to convert the value to String?

marc2332: Just pushed what I meant :)

marc2332 (approved):

PR #1227, opened 2025-06-10
Fix file references in README

Tynael:
Found a few more broken file references in the README.


marc2332 (approved):
I dont understand how I didnt notice these, weird, thanks again!

Issue #947, opened 2024-10-01
enhancement: Make scrollbar thumb have a minimum size


marc2332:
as more content the scroll view has, the smaller the thumb is, it would be nice if it had a minimum size

pahnin:
@marc2332 please have a look at #1228 when you have time, thanks

please do suggest changes

PR #1228, opened 2025-06-10
feat: default to a min value when setting size for scrollbar thumb

pahnin:
addressing https://github.com/marc2332/freya/issues/947

### DIFF
--- a/crates/components/src/scroll_views/scroll_thumb.rs
+++ b/crates/components/src/scroll_views/scroll_thumb.rs
@@ -26,6 +26,17 @@ enum ScrollThumbState {
     Hovering,
 }
 
+/// parse the n% string to n and return n% if n is greater than min
+/// returns min% if min is greater than n
+fn min_size(percentile_string: &str, min: i32) -> String {
+    let parsed_int = percentile_string
+        .trim_end_matches('%')
+        .parse::<i32>()
+        .unwrap_or(min);
+
+    format!("{}%", parsed_int.max(min))
+}
+
 /// Scroll thumb used for [`crate::ScrollView`] and [`crate::VirtualScrollView`].
 #[allow(non_snake_case)]
 pub fn ScrollThumb(
@@ -44,6 +55,9 @@ pub fn ScrollThumb(
         ScrollThumbState::Idle => theme.thumb_background,
         ScrollThumbState::Hovering => theme.hover_thumb_background,
     };
+
+    let width = min_size(&width, 5);
+    let height = min_size(&height, 5);
 
     rsx!(
         rect {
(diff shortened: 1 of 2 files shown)


pahnin:
I have made this change, but I am wondering if this should be configured somewhere else as a default or we should accept props

please feel free to comment and suggest I will make changes accordingly

pahnin:
hi @marc2332 thanks for taking time to review,

the way I have limited is to hardcoding to `5%` if the value is say `3%` or leaving it as is if it is say `7%`

do you think this 5 is acceptable or should it be more, also this applies both to horizontal scrolling and veritcal scrolling,
I think it might be more appropriate to have bigger value for horizontal scrolling? 

I will try to understand how the correlation between drag distance to scroll distance works today,
if you have any pointers on where I should be looking please let me know

thanks, I am really new to this space but at the same time very excited to work on this.. :)

pahnin:
hi @marc2332 I had a look at the code and have some questions
in https://github.com/marc2332/freya/blob/main/crates/components/src/scroll_views/virtual_scroll_view.rs#L309C17-L309C48
https://github.com/marc2332/freya/blob/main/crates/components/src/scroll_views/mod.rs#L70

from what I understand, when the thumb is dragged, we capture the position of the thumb and based on the value we calculate the position to scroll to, so it is based on the user input

if we change the logic to scroll less even though the user dragged `x` distance we have to reset the thumb back to the corresponding position once it is dragged?

marc2332:
> hi @marc2332 I had a look at the code and have some questions in [`main`/crates/components/src/scroll_views/virtual_scroll_view.rs#L309C17-L309C48](https://github.com/marc2332/freya/blob/main/crates/components/src/scroll_views/virtual_scroll_view.rs#L309C17-L309C48) [`main`/crates/components/src/scroll_views/mod.rs#L70](https://github.com/marc2332/freya/blob/main/crates/components/src/scroll_views/mod.rs#L70)
> 
> from what I understand, when the thumb is dragged, we capture the position of the thumb and based on the value we calculate the position to scroll to, so it is based on the user input
> 
> if we change the logic to scroll less even though the user dragged `x` distance we have to reset the thumb back to the corresponding position once it is dragged?

No need to reset, for every difference between the current clapped thumb size and the size it should have, it should scroll more distance

The change is purely mathematical and should not use any kind of "reset"/similar strategies

marc2332:
Sorry if I am a bit vague, its just that I have not spent enough time thinking about this, just enough to have an idea

pahnin:
no worries, it's a bit more clearer now, will try it out this week

pahnin:
hi @marc2332 looks like the scenario you explained is already covered, since the virtual scrollview is calculating the scrollposition based on the position of the thumb when the thumb is released rather than the size of the thumb

it takes the position and looks for the inner size of the content and calculates the position to scroll to

https://github.com/marc2332/freya/blob/main/crates/components/src/scroll_views/mod.rs#L70

I have also tested this using 

```
cargo run --example small_scroll_thumb
```

I think the next step is to decide on what is a reasonable min value is it acceptable if it is say `5%` or something else

let me know what you think :)

marc2332:

PR #1233, opened 2025-06-12
Consider 'grey' a valid syntax colour

Tynael:
Fixes the following error:

`thread '<unnamed>' panicked at C:\Users\<user>\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\freya-core-0.3.4\src\parsing.rs:26:17:
Failed to parse attribute 'Color' with value 'Text("grey")'`

`grey` is used mostly in the UK & Commonwealth countries
`gray` is used mostly in the US

I think both variants should be accepted 🤔


marc2332:
Hahaha, fair point!

marc2332:
Thanks!

PR #1292, opened 2025-07-04
add dev env flake

RobertasJ:


marc2332:
Is uploading the lock file also a thing? Or maybe it should be ignored?

RobertasJ:
Its like a cargo lockfile, its necessary to make the flake reproducible so as to not cause issues in the future, it can ofc be updated later.

marc2332:
> Its like a cargo lockfile, its necessary to make the flake reproducible so as to not cause issues in the future, it can ofc be updated later.

I actually don't push Cargo.lock here in freya (its recommended to dont do it in libraries). I guess for nix is fine maybe?

coolGi69:
For nix even libraries are meant to push their flake.lock. Without it, the flake becomes invalid and can't be used elsewhere.

PR #1310, opened 2025-07-23
add hot-patching support

RobertasJ:
Closes https://github.com/marc2332/freya/issues/1223

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,7 +83,7 @@ dioxus-devtools = { version = "0.7.0-alpha.3" }
 generational-box = { version = "0.7.0-alpha.3" }
 
 
-dioxus-clipboard = "0.2.0"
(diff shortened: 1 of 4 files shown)


alexculea:
This looks exciting! How how do we use it?

RobertasJ:
This thing has got an annoying bug as of now, but you just need to use dioxus-cli and pass the --hot-patch flag to it. Still trying to figure out the issue

RobertasJ:
> This looks exciting! How how do we use it?

you need to first use dioxus-cli with my pr that allows using only hot-patching: https://github.com/DioxusLabs/dioxus/pull/4475

and then use this freya in this pr.

to run the app you need to do `dx serve --hot-patch --hot-reload false`

marc2332:
Might be worth trying it in another time

PR #1311, opened 2025-07-23
nix documentation, added unstable rust devshell

RobertasJ:


coolGi69:
The `.envrc` probably shouldn't be in the `.gitignore` and should be added to the repo so that direnv works correctly.

RobertasJ:
I just made it ignored since i didnt want to put something i think is pretty easy to do normally anyways, the contents are just `use flake`.

marc2332 (approved):

PR #1316, opened 2025-07-25
feat: Ability to navigate using keyboard arrows only inside groups (e.g a Dropdown)

marc2332:
and Tab for the rest of nodes that are outside

Closes https://github.com/marc2332/freya/issues/1305

#### Old navigation:
- Tab = forward
- Shift + Tab = backward

#### New system:

To make it sort, "group members" basically mean those nodes that are members of the same group as the currently focused node

- Tab: forward excluding group members
- Shift + Tab: forward excluding group members
- Arrow Down: forward only in group members
- Arrow Up: backward only in group members

#### To think about
Horizontal arrows

### DIFF
--- a/crates/components/src/scroll_views/mod.rs
+++ b/crates/components/src/scroll_views/mod.rs
@@ -137,7 +137,7 @@ pub fn get_corrected_scroll_position(
 
 pub fn manage_key_event(
     key: &Key,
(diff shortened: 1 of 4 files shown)


marc2332:
Closing in favor of https://github.com/marc2332/freya/pull/1351

Issue #1308, opened 2025-07-22
docs: Add example showcasing stackable draggable elements


marc2332:
https://discord.com/channels/1015005816094478347/1397231814749196379

PR #1322, opened 2025-07-26
fix: Remove nodes from their old layer when their layer changes

marc2332:
Closes https://github.com/marc2332/freya/issues/1308


marc2332:

PR #1324, opened 2025-07-28
Wrapping layouts

dexterdy:


crates/core/src/values/wrap_content.rs
```diff
@@ -0,0 +1,16 @@
...
+    Parse,
+    ParseError,
+};
+
+impl Parse for WrapContent {
+    fn parse(value: &str) -> Result<Self, ParseError> {
+        Ok(match value {
+            "no-wrap" => WrapContent::NoWrap,
+            "wrap" => WrapContent::Wrap,
+            _ => WrapContent::NoWrap,
```
RobertasJ, line 13: this is incorrect, you need to return a parse error if the value type is incorrect

marc2332:
Closing in favor of https://github.com/marc2332/freya/pull/1408, unlike this PR, #1408 is more simple in both the feature and code implementation

Issue #44, opened 2022-11-09
enhancement: Multi window support


marc2332:

OlshaMB:
@marc2332 Is there temporary solution to this?
Is there a way to first create a window with then renderer ui?

marc2332:
> @marc2332 Is there temporary solution to this? Is there a way to first create a window with then renderer ui?

No, there is no solution for this right now

edfloreshz:
I believe [Dioxus](https://dioxuslabs.com/blog/release-030#multi-window-desktop-apps) supports this, is there any way to bring this to Freya?

marc2332:
> I believe [Dioxus](https://dioxuslabs.com/blog/release-030#multi-window-desktop-apps) supports this, is there any way to bring this to Freya?

It is probably possible but needs research and big refactors

PR #1325, opened 2025-07-29
feat: Multi Window

marc2332:
Closes https://github.com/marc2332/freya/issues/44

TO DO: 
- [x] ~~Figure out a decent API to pass props to the window root components, or maybe just move the state from launch config to window config~~
- [ ] ~~Research how to share automatic reactive state (not possible but maybe I can do something in dioxus-radio for instance)~~


Drigster:
Hi, can you please add ability for windows to close other windows, not only themselves. Returning a reference to the window or setting some type of ID when creating a window would be great. Thank you

PR #1326, opened 2025-07-30
feat: `rect` aspect ratio

RobertasJ:
adds aspect-ratio for any element. 

there is now a `aspect_ratio` attribute that has 3 values: "none", "fit" and "fill". none makes it act like normal, fit scales down the element when it doesnt fit inside the parent, fill always scales the element to fill the parent without overflowing.

### DIFF
--- a/crates/core/src/elements/image.rs
+++ b/crates/core/src/elements/image.rs
@@ -3,10 +3,19 @@ use freya_native_core::real_dom::NodeImmutable;
 
 use super::utils::ElementUtils;
 use crate::{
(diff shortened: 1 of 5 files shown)


RobertasJ:
Ive renamed min to Fill, and ive removed max because i cant think of a reason someone would use it. its not really a downgrade, and it also works on any element now

marc2332:
> Ive renamed min to Fill, and ive removed max because i cant think of a reason someone would use it. its not really a downgrade, and it also works on any element now

I have 4 examples using it so its definitely useful

marc2332:

Issue #1265, opened 2025-06-18
docs: Example on to send OS notifications using notify-rust


marc2332:
https://github.com/hoodie/notify-rust

mistermaxis:
Hello Marc. I was in the process of trying to learn this framework, which I think is very nice even if I'm struggling a bit with some things. In any case, I created an example boilerplate project using some basic notifications from notify-rust. If I could, I would like to make a pull request with it.
Regards.

marc2332:
> Hello Marc. I was in the process of trying to learn this framework, which I think is very nice even if I'm struggling a bit with some things. In any case, I created an example boilerplate project using some basic notifications from notify-rust. If I could, I would like to make a pull request with it. Regards.

Hey Maximiliano! That would be very much appreciated, feel free to open a PR anytime 😄

mistermaxis:
Excellent! Thank you very much 😀

mistermaxis:
Hi Marc. I created a pull request with the examples for notify rust. Let me know if there is anything you would like me to change.
Also, I would like to ask you some basic questions that came up while working on this:
- Is there a way to use signals across threads?
- What is the proper way to do media queries?
- Is there a built-in way to have a menu/sidebar collapse (like a snackbar or an accordion?

marc2332:
> * Is there a way to use signals across threads?

There is `use_signal_sync`, but if you are doing async and not actual parallelism I strongly suggest you stick to `use_signal` and instead the async APIs from dioxus (See more here https://docs.rs/freya/latest/freya/_docs/async_tasks/index.html)

> * What is the proper way to do media queries?

I don't have any doc/or fancy API to do so atm, the closest thing would be to use `use_platform_information` as shown in https://github.com/marc2332/freya/blob/main/examples/window_size.rs

> * Is there a built-in way to have a menu/sidebar collapse (like a snackbar or an accordion?

I have a Sidebar component (as shown in https://github.com/marc2332/freya/blob/d9dd2d3f4d362e360fe1f56976b6664e5b646010/examples/animated_sidebar.rs#L127) but it has no concept of being collapsed, but it sounds like a good idea, just created an issue for it https://github.com/marc2332/freya/issues/1331.

That said, I do also have a `Snackbar` (https://github.com/marc2332/freya/blob/main/examples/snackbar.rs) and `Accordion` (https://github.com/marc2332/freya/blob/main/examples/accordion.rs) components

marc2332:
Not necessary IMO

PR #1330, opened 2025-08-03
Notify rust examples

mistermaxis:
Closes https://github.com/marc2332/freya/issues/1265

#### Added example file for notify-rust examples:
- Simple example
- Simple example (async)
- Minimal example
- Example with actions (options)
- Example with callback function on close
- Example with callback function and reason (for closing the notification) message


examples/notify_rust.rs
```diff
@@ -0,0 +1,432 @@
...
+)]
+
+use std::thread;
+
+use freya::prelude::*;
+use freya_router::prelude::*;
+use notify_rust::{Hint, Notification};
+
+fn main() {
+    launch_with_props(app, "Router Example", (550.0, 400.0));
```
marc2332, line 13: Not your fault, I changed the name of this method yesterday :D
```suggestion
    launch_with_params(app, "Router Example", (550.0, 400.0));
```

marc2332 (changes requested):
Seems like the CI is failing, maybe some methods are not available in MacOS?

examples/notify_rust.rs
mistermaxis: #### Yeah, I noticed that was changed after I made the PR hehe.
#### So, I'm not sure what's the best way to resolve this. At first I tried using #[cfg(target_os = ...)] rules like in the original notify-rust examples to include the problematic examples conditionally, but then errors with the formatting started showing up, so in the end I just removed all the examples that were Linux/Unix specific.
#### Now, there's an error on the Ubuntu build about a "blocking" module in the zbus crate. That, I don't know what to do about :sweat_smile:

mistermaxis: Excellent! Thank you so much. Let me know if there is anything I can help with.

marc2332:
@mistermaxis Seems like the CI is still failing 👀

mistermaxis:
> @mistermaxis Seems like the CI is still failing 👀

Yes, I haven't been able to figure out how to remove these errors, which come from a different crate (zbus) and doesn't mention anything about any change in that module (blocking-api) in recent versions. Also, dependency versions are not something I'm all that familiar with. I will keep investigating, probably consult their community. At least now it fails on my machine too, so that should help me make progress more easily.

marc2332:
Closing due to inactivity

Issue #1176, opened 2025-05-22
bug: absolute positioning only takes initial size into account


chrrs:
When using `position: "absolute"` with a `position_bottom` value, it's only correctly translated if the node has a size independent from it's children. I haven't tested this with `position_right`, but I'd imagine it would be similar.

```rust
fn app() -> Element {
    rsx!(
        rect {
            background: "blue",
            width: "100",
            height: "100",

            rect {
                background: "green",
                width: "40",
                height: "40",

                position: "absolute",
                position_bottom: "0",
                position_left: "0",
            }

            rect {
                background: "yellow",
                width: "40",
                height: "auto",

                position: "absolute",
                position_bottom: "0",
                position_left: "50",

                rect { height: "40" }
            }
        }
    )
}
```

![Image](https://github.com/user-attachments/assets/b4091318-022c-442f-8018-afbe7a827c3b)

I would expect the yellow square here to be inside the blue square, not outside it.

chrrs:
Ah, that makes sense! I'll work around it for now then.

PR #1342, opened 2025-08-27
Fix/absolute inner size

dexterdy:
A fix for the incorrect placement of inner sized absolutely positioned elements, by redetermining the origin after the inner size was calculated in the initial phase


marc2332:
Closing as I did not like the solution back then, also, the code is outdated. Issue -> https://github.com/marc2332/freya/issues/1176

Issue #1348, opened 2025-10-29
text editor hotkeys (paste, undo, etc) do not respect user keyboard layout


justDeeevin:
I'm a colemak-dh user and I noticed that, in the text editor examples, ctrl+x causes an undo, rather than ctrl-z. This is because the x key in colemak-dh is in the place where z would be in qwerty, and onkeydown uses keycodes (that is, key position rather than the character itself). Notably, typing text into the editors does work as expected, using the correct characters for my keyboard layout.

I don't know if the intent for onkeydown is to report the keycode rather than the character itself, but regardless of whether this necessitates a change in onkeydown's behavior, the text editors' behaviors, or perhaps entirely new functionality for input handling itself, users will certainly expect text input hotkeys to respect their keyboard layouts, so that should be fixed.

marc2332:
Mmmm, I see, I will check this out!

marc2332:
Hey @justDeeevin Can you try with this branch? https://github.com/marc2332/freya/pull/1351 I pushed a fix just now that removes the use of any hardware key. If you can't or don't want, it's fine! Seems to work fine from my own testing.

Issue #1343, opened 2025-08-28
Add support for 8-digit hex colors


Tynael:
If we're using an 8-digit hex color, Freya will panic.

Example:
```
Link {
    to: "https://example.com",
    label {
        color: "#872b2bff",
        "Text",
    }
}
```
```
thread '<unnamed>' panicked at C:\Users\<user>\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\freya-core-0.3.4\src\parsing.rs:26:17:
Failed to parse attribute 'Color' with value 'Text("#872b2bff")'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

kyteidev:
the PR to add support for 8 character HEX (#1205) was already merged but I think it will be released in version 0.4.0

Issue #1332, opened 2025-08-10
enhancement: Add feature to expose skia-safe APIs


marc2332:
This way using `use_canvas` gets easier

Issue #1319, opened 2025-07-25
Make theme overwriting easier


Drigster:
Using theme for changing single values is not very convinient.
You need to define a theme:
```
Button {
    theme: theme_with!(ButtonTheme {
        margin: "8",
    }),
    label {
        "Button"
    }
}
```
Instead of using the value directly:
```
Button {
    margin: "8",
    label {
        "Button"
    }
}
```

PR #1351, opened 2025-10-30
A new begining :sparkles:

marc2332:
> :warning: Warning: This is still work in progress, there might be missing features or bugs.

> **I will continue to update this PR (and this PR description) until I consider it stable and ready**

TLDR: Freya (v0.4) now has its own reactive and component model, and it no longer relies on Dioxus for it, although the new system is heavily inspired by it, but with some key differences:
- There is no `rsx!()` macro
- Elements are user-land
- Attributes are fully typed
- Components are now any data type that implements the `Render` trait, so the "props" are now the data type itself. 

But it still has:
- Hooks (but e.g `use_signal` is now `use_state`)
- Callback event handlers
- And generally just a very similar state/async tasks management

Project changes:
- Less deps (Will share more when I can)
- Forked crates (I created them though!):
   - https://github.com/dioxus-community/dioxus-radio
   - https://github.com/dioxus-community/dioxus-i18n
   - https://github.com/dioxus-community/dioxus-clipboard
- New and splitted crates:
   - `freya-icons`: An Icons library (Only supports Lucide at the moment) 
   - `pathgraph`: Path-based graph (every item in a path is an index indicating the location of a item's value) 
   - `freya-performance-plugin`: Contains the Performance overlay Plugin
   - `freya-edit`: Contains all the text editing APIs, like `TextEditor`, `use_editable`, etc
   - `freya-animation`: Contains the `use_animation` hook and all its related APIs
- Removed crates:
   - `freya-hooks`: The  equivalent APIs are now included in `freya-core`, `freya-animation`, `freya-edit` and `freya-components`
   - `freya-native-core` & `freya-native-core-macro`: This was the glue used to connect with Dioxus, so its not needed anymore.
   - `freya-elements`: This used to contain the elements and attributes definitions, the equivalent is now included in `freya-core`.
   - `freya-sdk`: Generic utilities for Freya apps, like integrations with Tokio.
- Uses [just](https://github.com/casey/just) to run development scripts
- Uses [hotpath](https://github.com/pawurb/hotpath) to debug and improve performance

- Closes https://github.com/marc2332/freya/issues/1343
- Closes https://github.com/marc2332/freya/issues/1332
- Closes https://github.com/marc2332/freya/issues/1327
- Closes https://github.com/marc2332/freya/issues/1319
- Closes https://github.com/marc2332/freya/issues/1308
- Closes https://github.com/marc2332/freya/issues/1305
- Closes https://github.com/marc2332/freya/issues/1304
- Closes https://github.com/marc2332/freya/issues/1290
- Closes https://github.com/marc2332/freya/issues/1267
- Closes https://github.com/marc2332/freya/issues/1264
- Closes https://github.com/marc2332/freya/issues/1198
- Closes https://github.com/marc2332/freya/issues/947
- Closes https://github.com/marc2332/freya/issues/835
- Closes https://github.com/marc2332/freya/issues/1348
- Closes https://github.com/marc2332/freya/issues/860
- Closes https://github.com/marc2332/freya/issues/1051
- Closes https://github.com/marc2332/freya/issues/886
- Closes https://github.com/marc2332/freya/issues/995
- Closes https://github.com/marc2332/freya/issues/909
- Closes https://github.com/marc2332/freya/issues/347
- Closes https://github.com/marc2332/freya/issues/552
- Closes https://github.com/marc2332/freya/issues/951

- Supersedes https://github.com/marc2332/freya/pull/1259
- Supersedes https://github.com/marc2332/freya/pull/1306
- Supersedes https://github.com/marc2332/freya/pull/1322
- Might supersede https://github.com/marc2332/freya/pull/1298
- Supersedes https://github.com/marc2332/freya/pull/1210
- Supersedes https://github.com/marc2332/freya/pull/1328
- Supersedes https://github.com/marc2332/freya/pull/1326

The goal is to reach 100% feature (with some deprecations), docs examples and tests parity with `main`.

Code might be a bit ugly in some parts! I will improve it.

WIP.

Notes for me:
- [x] idea: explorer some kind of reactive state adapter for async channels
- [x] add: `.layer` for image, label and paragraph
- [x] to do: Allow ImageViewer to accept children
- [ ] to do: Downscaling optimizations for ImageViewer
- [x] to do: Example equivalent of https://github.com/marc2332/freya/blob/main/examples/render_canvas.rs
- [x] to do: Custom text cache
   - [x] to do: Add tests
- [x] to do: Pretty hook errors
- [x] to do: Allow closing created windows
- [x] to do: https://github.com/marc2332/freya/discussions/1357
- [ ] to do: https://github.com/marc2332/freya/discussions/1358
- [x] to do: Allow setting the initial position for `Draggable`
- [ ] to do: Optimize Skia cache?
- [x] to do: port https://github.com/marc2332/freya/pull/1356
- [x] to do: allow creating radio station in main function

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 **Freya** is a **cross-platform and non-web** GUI library for Rust powered by 🎨 [Skia](https://skia.org/).
 
-> :warning: I recently rewrote a huge percentage of Freya in https://github.com/marc2332/freya/pull/1351, so the `main` branchd differs a lot from the latest stable release.
+> :warning: I recently rewrote a huge percentage of Freya in https://github.com/marc2332/freya/pull/1351, so the `main` branch differs a lot from the latest stable release.
 
 #### Counter example
 <table>


marc2332:
Seems like the Rust CI is failing because taplo needs WSL on Windows? Makes no sense! I need to investigate this

Saul-BT:
LGTM! 🔥

AlexMayol (approved):
Deploy first, test later :shipit:

Saul-BT (approved):

marc2332:
Gotta use a shared font for text editing tests, otherwise each OS will do things differently

marc2332:
idea: explorer some kind of reactive state adapter for async channels

tekacs:
@marc2332 I hope you don't mind me asking – I'm curious if there's any reason behind this change away from Dioxus, other than to make it easier to resolve some issues that arise from the two layers together!

marc2332:
> @marc2332 I hope you don't mind me asking – I'm curious if there's any reason behind this change away from Dioxus, other than to make it easier to resolve some issues that arise from the two layers together!

No problem! I was going to mention them once I released this new version of Freya anyway. Some of the reasons area:

- No `rsx!()` macro -> better autocomplete, sounds silly but is a huge difference
- Elements are user-land -> allows users to hook into the lower parts of Freya by creating new elements
- Attributes are fully typed -> No magic strings -> better API discoverability and more performance as Freya doesnt have to parse these values
- Dioxus as a whole was kind of a black box for me, for certain things I didnt feel like I was in control, now that I control it all I can make better API decision.
Also, Dioxus has a very slow release pace and usually has huge changes, these are not easy to track nor adapt sometimes.
  - Also, in the past there has been quite a few memory leaks (that I reported to them) but I guess I feel like these leaks might come again or new might arise at any moment and I would not notice them right away, this makes me want to depend less on Dioxus, these reports might take time to debug, fix and release. If I am going to have a memory leak at least I want to be the one to fix it as that way I can release it asap.
- There are a few other API decisions like the fact that Components by default get a reference to their data (aka props) rather than an Owned version like Dioxus, this is because Dioxus clones the props in every run of any component, even though you might not even need to consume those. For the sake of allowing that in Freya I added `RenderOwned` too, but the way to go remains being `Render` nonetheless.
- Less complexity, by having control over all the stack now there is just less code to glue everything, there are fewer abstractions (e.g previously the nodes tree was a completely different crate to where the other trees or related data, were stored), in the other hand now almost all tree-related logic lives now in `tree.rs`
   - And at the same time, there is also less runtime complexity, as I mentioned with Dioxus attributes were filled with string values, Dioxus can technically support other values like bools, floats, numbers and `Rc<any>`, but none of these were as flexible and easy to use as Strings, yet these strings come with a runtime parsing and allocation cost. A cost that is not worth doing, its like a wall that that should not be there, we are in Rust, there is no need to be using Strings as an API transport, just let me pass the value directly!
- The component model of Dioxus does not scale in my opinion, it tries to escape from what makes Rust good, and this is for example having custom constructors and Traits, Dioxus components do not work like this, instead you are left with having to specify default props values with macros and havint to call the components like `CoolComponent { value: 1 }` rather than e.g `CoolComponent::new(1)`, all this unnecesary complexity in Dioxus components makes it impossible to expend Components, e.g in Dioxus you cannot implement a trait for a Dioxus component (or element even) to add a new `prop`, a component can only take the props it was defined with. In the other hand in Freya you can do something like:
```rs
#[derive(PartialEq, Default)]
struct CoolComponent(pub u8);

trait RandomTrait: where Self: Sized {
   fn with_123(mut self) -> Self;
}

impl RandomTrait for CoolComponent {
   fn with_123(mut self) -> Self {
       self.0 = 123;
       self
   }
}

impl Render for CoolComponent {
   fn render(&self) -> Element {
      rect().into()
   }
}

// So that you can call it like

return CoolComponent::default().with_123().into();
```

I want to clarify that Dioxus does have things that I like, like Hooks, Copy state management and the async model. I modeled the new freya equivalents after Dioxus.

This is all I could think of right now, I will probably write it in a more concise and clear way once released :)

tekacs:
Thanks for taking the time to post this @marc2332! Really helpful to hear!

I use Dioxus for my app (very happily!) but it'll be nice to _also_ have the API that you're putting together (and continuing to use some patterns from Dioxus around hooks and generational box) for a different set of circumstances.

Thanks for putting this together and looking forward to it!

marc2332:
This PR has a lot of noise commits, so I will squash it

PR #1353, opened 2025-11-01
feat: text editing improvements

tropicaaal:
- Implements Home/End navigation to jump to start/end of line and start/end of buffer
- Fixes Left/Right arrow with selection to jump to start/end of selection.
- Improves vertical navigation by remembering the cursor's column during the last horizontal navigation action.

Some of this behavior could probably be cleaned up if we cleaned up the API for moving a cursor (I have a few ideas here). We still need to implement grapheme-level navigation when holding `Ctrl` + left/right. Have left that as a todo since we'll need to bring in `unicode_segmentation` for that.

### DIFF
--- a/crates/freya-edit/src/config.rs
+++ b/crates/freya-edit/src/config.rs
@@ -1,6 +1,6 @@
 #[derive(Clone, Copy, PartialEq, Debug)]
 pub struct EditableConfig {
-    pub(crate) identation: u8,
+    pub(crate) indentation: u8,
     pub(crate) allow_tabs: bool,
     pub(crate) allow_changes: bool,
     pub(crate) allow_clipboard: bool,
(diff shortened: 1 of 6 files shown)


tropicaaal:
A possible idea to clean up this behavior would be modify `cursor_left`/`cursor_right` to take a `CursorMovement` enum.

e.g:
```rs
enum CursorMovement {
    /// Move one character left/right.
    Character,

    /// Move one word left/right.
    Word,

    /// Move to start/end of line.
    Line,

    /// Move to start/end of paragraph.
    Paragraph,

    /// Move to start/end of buffer.
    Buffer,
}
```

which specifies the granularity of the movement.

marc2332:
Oh, seems like GitHub closed the PR when the `a-new-begining` branch was deleted

marc2332:
I didnt mean to close this

PR #1370, opened 2025-12-20
chore(edit): Refactor `freya-edit` to use a selection-based cursor

marc2332:
Simplifies the internal data structures to avoid maintaining the cursor and the selected text separately, instead only 1 exists at a time.


marc2332:
I will split this in 2 PRs

PR #1396, opened 2025-12-28
fix(core): Properly drop node to scope mapping

marc2332:
This fixes a leak where non-scope-owner nodes were not being removed from the node-to-scope mapping, at the same time I think it's also somewhat faster than before, but I did not run the numbers, so I am not entirely sure.

to do: Add test.


marc2332:
Okay... forget about being faster, its slower

marc2332:
Nice! Performance is good again.

PR #1450, opened 2026-01-14
Improve the `MaybeExt` trait.

bowenxuuu:
Adding some methods like `maybe_else`, `maybe_some`, and `maybe_none` to the `MaybeExt` trait.

### DIFF
--- a/crates/freya-core/src/elements/extensions.rs
+++ b/crates/freya-core/src/elements/extensions.rs
@@ -846,16 +846,34 @@ pub trait MaybeExt
 where
     Self: Sized,
 {
-    fn maybe(self, bool: impl Into<bool>, then: impl FnOnce(Self) -> Self) -> Self {
-        if bool.into() { then(self) } else { self }
+    /// Imperatively modify self with the given closure.
+    fn map<U>(self, f: impl FnOnce(Self) -> U) -> U
+    {
+        f(self)
     }
 
-    fn map<T>(self, data: Option<T>, then: impl FnOnce(Self, T) -> Self) -> Self {
-        if let Some(data) = data {
-            then(self, data)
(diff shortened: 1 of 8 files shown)


crates/freya-core/src/elements/extensions.rs
```diff
@@ -846,17 +846,35 @@ pub trait MaybeExt
...
-    fn map<T>(self, data: Option<T>, then: impl FnOnce(Self, T) -> Self) -> Self {
-        if let Some(data) = data {
-            then(self, data)
+    /// Conditionally modify self with the given closure.
+    fn maybe_else(self, condition: impl Into<bool>, then: impl FnOnce(Self) -> Self, else_fn: impl FnOnce(Self) -> Self) -> Self {
+        if condition.into() { then(self) } else { else_fn(self) }
+    }
+
+    /// Conditionally unwrap and modify self with the given closure, if the given option is Some.
+    fn maybe_some<T>(self, option: Option<T>, then: impl FnOnce(Self, T) -> Self) -> Self {
```
marc2332: Being called `map` like before feels more rusty

crates/freya-core/src/elements/extensions.rs
```diff
@@ -846,17 +846,35 @@ pub trait MaybeExt
 where
     Self: Sized,
 {
-    fn maybe(self, bool: impl Into<bool>, then: impl FnOnce(Self) -> Self) -> Self {
-        if bool.into() { then(self) } else { self }
+    /// Imperatively modify self with the given closure.
+    fn map<U>(self, f: impl FnOnce(Self) -> U) -> U
```
marc2332: Some idea
```suggestion
    fn with<U>(self, f: impl FnOnce(Self) -> U) -> U
```

PR #1583, opened 2026-03-03
Fix pointer enter/leave hit-testing for rects at large coordinates

ghostnumber7:
# Fix Pointer Enter/Leave at Large Coordinates

## Summary

This PR fixes a hit-testing precision issue where `pointer_enter` / `pointer_leave` could fail for elements positioned far from the origin (reported around `2048px+`, observed in WSL2 scenarios).

## Problem

When a `rect` is located at large absolute coordinates, pointer hover transitions may stop working:

- `on_pointer_enter` is not emitted when moving inside the element.
- `on_pointer_leave` is not emitted when leaving the element.

## Root Cause

`RectElement::is_point_inside` used `SkRRect::contains` with an epsilon-sized absolute `SkRect` around the cursor:

- `SkRect::new(x, y, x + 0.0001, y + 0.0001)`

At high coordinate magnitudes, `f32` precision can make `x + 0.0001` (or `y + 0.0001`) collapse to `x` (`y`) after rounding, producing unstable containment checks.

In short: the hit-test used very small deltas in a high-magnitude `f32` space.

## Solution

Keep the same hit-testing logic, but make it precision-safe:

1. Build the rounded rect in local space (`origin = (0, 0)`, same size/radius).
2. Translate cursor to local coordinates:
   - `local_x = cursor.x - area.min_x()`
   - `local_y = cursor.y - area.min_y()`
3. Build the tiny test rect using the next representable `f32` value (1 ULP) per axis instead of a fixed `0.0001` delta.
4. Call `contains` with that local, ULP-based rect.

This preserves behavior while avoiding large-magnitude floating precision issues in both offset-heavy and size-heavy cases.

## Why this approach

- Minimal and local fix.
- No semantic change to rounded-corner hit-testing.
- No allocations; negligible runtime impact (constant-time arithmetic only).
- Avoids broad architectural changes (Skia/layout APIs are `f32`-based anyway).

### 1) `pointer_enter_leave_at_large_coordinates`

Validates that enter/leave events work for a target element positioned around `y = 3000`.

### 2) `large_coordinate_hover_does_not_trigger_other_elements`

Validates no cross-triggering:

- A hoverable element at `0,0`
- A hoverable element at `y=3000`

Hovering one must not emit handlers for the other.

### 3) `pointer_enter_on_huge_rect_far_local_coordinate`

Validates that a huge rect still emits `pointer_enter` at far in-bounds local coordinates (e.g. `x = 5000`) where fixed epsilon would otherwise collapse.

### 4) `pointer_enter_at_extreme_offset_coordinate`

Validates behavior at extreme absolute offsets (e.g. `y = 1_000_000_000_000`):

- This case fails without local-coordinate rounded-rect construction.
- It passes with local-coordinate construction + ULP-based probe rect.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -131,8 +131,8 @@ ash = "^0.38"
 ash-window = "0.13.0"
 
 # accessibility
-accesskit = "0.23.0"
-accesskit_winit = "0.31.0"
+accesskit = "0.24.0"
+accesskit_winit = "0.32.0"
 
 # Performance
 hotpath = "0.10"
@@ -177,7 +177,7 @@ skia-safe = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
 
-rand = "0.9"
+rand = "0.10"
 
 async-io = { workspace = true }
 futures-lite = { workspace = true }
--- a/crates/freya-components/src/image_viewer.rs
+++ b/crates/freya-components/src/image_viewer.rs
@@ -181,6 +181,7 @@ pub struct ImageViewer {
     image_data: ImageData,
     accessibility: AccessibilityData,
     effect: EffectData,
+    corner_radius: Option<CornerRadius>,
 
     children: Vec<Element>,
 
@@ -195,6 +196,7 @@ impl ImageViewer {
             image_data: ImageData::default(),
             accessibility: AccessibilityData::default(),
             effect: EffectData::default(),
+            corner_radius: None,
             children: Vec::new(),
             key: DiffKey::None,
         }
@@ -237,6 +239,13 @@ impl ChildrenExt for ImageViewer {
 impl EffectExt for ImageViewer {
     fn get_effect(&mut self) -> &mut EffectData {
         &mut self.effect
+    }
+}
+
+impl ImageViewer {
+    pub fn corner_radius(mut self, corner_radius: impl Into<CornerRadius>) -> Self {
+        self.corner_radius = Some(corner_radius.into());
+        self
     }
 }
 
@@ -299,6 +308,9 @@ impl Component for ImageViewer {
                     .image_data(self.image_data.clone())
                     .effect(self.effect.clone())
                     .children(self.children.clone())
+                    .map(self.corner_radius, |img, corner_radius| {
+                        img.corner_radius(corner_radius)
+                    })
                     .into_element()
             }
             Asset::Pending | Asset::Loading => rect()
--- a/crates/freya-components/src/input.rs
+++ b/crates/freya-components/src/input.rs
@@ -500,7 +500,7 @@ impl Component for Input {
                 (theme_colors.background, None, None)
             };
 
-        let border = if focus_status() == FocusStatus::Keyboard {
+        let border = if focus_status().is_focused() {
             Border::new()
                 .fill(theme_colors.focus_border_fill)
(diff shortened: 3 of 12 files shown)


marc2332:
@ghostnumber7 This was clearly AI generated. Where did you have this issue? I never had it

ghostnumber7:
> @ghostnumber7 This was clearly AI generated. Where did you have this issue? I never had it

It is AI generated indeed.
The issue arouse because I had a ui with a hover element way to the right of it and my monitor is 3400px wide.
Then I added a hover on an element to see what I was doing wrong and I noticed tracking on rect would stop working at exactly 2048px from the start, but global mouse move in the whole window would not have any issue.
Debugging for a while I found that to be the cause.
The tests will fail with current code so you can see it replicate in those conditions (the huge many millions width is just a "still works with absurd things" but the others are normal cases)

The whole problem is that for f32 a 2048.0 + 0.0001 == 2048.0 so the small rect used for testing is 0 px wide.

The larger the number the worst accuracy gets. I considered replacing the 0.0001 with f32::EPSILON but that is just the same as f32::next_up(1.0) - 1.0 which would still be the same thing. 

By comparing in local domain (subtracting the left / top offset) it reduces the number so that works on its own with elements anywhere for up to 2048px wide, but for large elements the f32::next_up() is required in the far side of them anyway.

marc2332:
> 2048.0 + 0.0001 == 2048.0

Not really, it's not the same

Also, why 2048? Where is this number coming from?

Why does the position matter?

marc2332:
Your changes look interesting nonetheless. I tried used epsilon some time ago but had no luck

ghostnumber7:
> Also, why 2048? Where is this number coming from?

It's due to how f32 works look [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=38b1da325f6a5df8bc95253c9836bc20). The image shows the playground output 

image

marc2332:
> > Also, why 2048? Where is this number coming from?
> 
> It's due to how f32 works look [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=38b1da325f6a5df8bc95253c9836bc20). The image shows the playground output
> image

Right... 🤔

ghostnumber7:
And [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=59e1cc85cd6bfaec3ebc542759a64516) is why the position matters

image

marc2332:
Will merge asap!

ghostnumber7:
> Will merge asap!

Thanks ... Loving the library by the way :) bona feina

marc2332:
Just realized your commits @ghostnumber7 are not signed. If you are up for it you gran grab this PRs commits (including mines) and rebase them while signing them. I can't seem to be able to push either. Having you created this PR from the `main` branch complicates things a lot. I can also close this PR and open it myself with your changes

ghostnumber7:
I don't mind either. The credit is not something I care for, but I can do it in a few if you want me to.

marc2332:
> I don't mind either. The credit is not something I care for, but I can do it in a few if you want me to.

Yeah, if you can, it would be nice. No hurry tho

ghostnumber7:
Made a bit of a mess but it's not hopefully correct.

marc2332:
> Made a bit of a mess but it's not hopefully correct.

Happens to the best of us! Thanks :)

Issue #1469, opened 2026-01-18
enhancement: Touch support for scrolling


marc2332:
Add support for the scroll views components to scroll by dragging with a finger

PR #1590, opened 2026-03-04
Add experimental Android support

SparkyTD:
This PR introduces some fixes and changes that are required for Android support.

## Summary of changes
- Add support for user-specified custom `EventLoopBuilder` at window creation. This will be required in order to create a custom event loop using `EventLoop::with_android_app(app: AndroidApp)`.
- ~~Fix a hit-testing issue in is_point_inside() by multiplying the hit rect size by the scale factor. This does not pose an issue on Desktop, but breaks hit-testing completely on the high scaling factor on most Android device. Likely due to internal floating point or rounding error in Skia.~~ (Already fixed by https://github.com/marc2332/freya/pull/1583)
- Properly integrate touch events into the pointer press and global event handling logic.
- Recreate the app window rendering context when the app is "resumed", i.e. when the user moves the app into the background, and then re-opens it on Android.
- ~~Handle global touch capture events on the `Slider` component~~ (Already fixed by https://github.com/marc2332/freya/pull/1591)
- Add support for ScrollView and VirtualScrollView touch scrolling on Android
- Add an example project that can be built for both Android and Desktop

## Currently known issues
- [ ] No soft keyboard / IME support

## Issues
- Closes https://github.com/marc2332/freya/issues/1469
- Closes https://github.com/marc2332/freya/issues/1691

### DIFF
--- a/crates/freya-winit/src/drivers/gl.rs
+++ b/crates/freya-winit/src/drivers/gl.rs
@@ -223,7 +223,9 @@ impl OpenGLDriver {
 
         window.pre_present_notify();
         self.gr_context.flush_submit_and_sync_cpu();
-        self.gl_surface.swap_buffers(&self.gl_context).unwrap();
+        if let Err(error) = self.gl_surface.swap_buffers(&self.gl_context) {
+            tracing::error!("Failed to swap buffers: {:?}", error);
+        }
     }
 
     pub fn resize(&mut self, size: PhysicalSize<u32>) {


marc2332:
Merged https://github.com/marc2332/freya/pull/1591, you should be able to revert those changes related to touch events :)

crates/freya-winit/src/drivers/gl.rs
```diff
@@ -223,7 +223,9 @@ impl OpenGLDriver {
 
         window.pre_present_notify();
         self.gr_context.flush_submit_and_sync_cpu();
-        self.gl_surface.swap_buffers(&self.gl_context).unwrap();
```
marc2332: Is a silent error really the way to go? Wouldn't that be a bit of a fatal error?

marc2332 (changes requested):

crates/freya-winit/src/drivers/gl.rs
marc2332: or is it recoverable / fail to ignore?

marc2332:
I pushed a new event loop  customization api, see https://github.com/marc2332/freya/blob/main/examples/feature_custom_event_loop.rs

marc2332:
Opened the same PR in https://github.com/marc2332/freya/pull/1698 as GitHub didnt let me push here

marc2332:
> Opened the same PR in #1698 as GitHub didnt let me push here

merged!

PR #1644, opened 2026-03-15
feat: Flexible launch methods

marc2332:
Kinda reverts https://github.com/marc2332/freya/pull/1637 as I realized it was not flexible enough. I also want access to the event loop.


marc2332:
Actually, let's just allow the user to pass its own eventloop.

Issue #1691, opened 2026-03-24
bug: Touch event not properly recognized by `on_press` event handler


MierenManz:
the `on_press` event handler only recognizes touch events if a mouse event has trigged it previously.

in the repro below (stolen from counter.rs) the Increase button does not increase the count when a touchscreen is used. It only works when the mouse has been used previously.

For example:
1. MOUSE: click on "Increase" and the counter goes up
2. TOUCH: click on "Increase" and the counter goes up
3. MOUSE: click on "Decrease" and the counter goes down.
4. TOUCH: click on "Increase" and the counter does not go up.

The fourth action should be the same as second action.

So far I have tried to see if it was taking the cursor position into account, but hovering over the button does not change the behavior.

```rs
#![cfg_attr(
    all(not(debug_assertions), target_os = "windows"),
    windows_subsystem = "windows"
)]

use freya::prelude::*;

fn main() {
    launch(LaunchConfig::new().with_window(WindowConfig::new(app).with_size(500., 450.)))
}

fn app() -> impl IntoElement {
    let mut count = use_state(|| 4);

    let counter = rect()
        .width(Size::fill())
        .height(Size::percent(50.))
        .center()
        .color((255, 255, 255))
        .background((15, 163, 242))
        .font_weight(FontWeight::BOLD)
        .font_size(75.)
        .shadow((0., 4., 20., 4., (0, 0, 0, 80)))
        .child(count.read().to_string());

    let actions = rect()
        .horizontal()
        .width(Size::fill())
        .height(Size::percent(50.))
        .center()
        .spacing(8.0)
        .child(
            Button::new()
                .on_press(move |event: Event<PressEventData>| {
                    let evt_type = match event.data() {
                        PressEventData::Touch(_) => "Touch",
                        PressEventData::Mouse(_) => "Mouse",
                        PressEventData::Keyboard(_) => "Keyboard",
                    };

                    println!("On Press triggered by {}", evt_type);
                    *count.write() += 1;
                })
                .child("Increase"),
        )
        .child(
            Button::new()
                .on_press(move |_| {
                    println!("Decreased has been pressed");
                    *count.write() -= 1;
                })
                .child("Decrease"),
        );

    rect().child(counter).child(actions)
}
```

marc2332:
Might get fixed by https://github.com/marc2332/freya/pull/1590

PR #1692, opened 2026-03-24
fix(core): Proper touch released event

marc2332:
Extracted from https://github.com/marc2332/freya/pull/1590

Closes https://github.com/marc2332/freya/issues/1691


MierenManz:
Fix seems to work mark :D

Issue #1703, opened 2026-03-25
Text overflowing parent when using max_width


henri2h:
Hi,
when using max_width on rect, child will overflow its parent.
Is it expected?

```rust
fn app() -> impl IntoElement {
    rect()
        .background(Color::BLUE)
        .max_width(Size::percent(72.))
        .child(label().text("lorem ipsum and so on, this text will overflow the parent rect"))
}
```

<img width="875" height="107" alt="Image" src="https://github.com/user-attachments/assets/f0bd8362-475c-484c-b9df-0753a1869d8a" />

marc2332:
max sizing are abit special, they dont work like normal sizes.

You probably want something like this instead:

```rs
fn app() -> impl IntoElement {
    rect()
        .background(Color::BLUE)
        .width(Size::percent(72.))
        .child(
            label()
                .width(Size::fill())
                .max_lines(1)
                .text_overflow(TextOverflow::Ellipsis)
                .text("lorem ipsum and so on, this text will overflow the parent rect"),
        )
}

```

henri2h:
Thanks, the idea I had when using the max_width is to let the parent shrink to the size of the children while making
sure the width wouldn't exceed a specific percentage (like the chat bubbles in an instant messaging app).
 I guess it's not possible to do then?

marc2332:
yes @henri2h, I opened a fix: https://github.com/marc2332/freya/pull/1704, with that PR the following will work:

```rust
fn app() -> impl IntoElement {
    rect()
        .background(Color::BLUE)
        .max_width(Size::percent(72.))
        .child(
            label()
                .text_overflow(TextOverflow::Ellipsis)
                .text("lorem ipsum and so on, this text will overflow the parent rect"),
        )
}
```

marc2332:
https://github.com/user-attachments/assets/8bf6d48e-fd25-42a7-a6fa-0333d73335db

henri2h:
Works great thanks a lot!!

PR #1704, opened 2026-03-25
fix(torin): Use parent area after inner sized min max measurement

marc2332:
Closes https://github.com/marc2332/freya/issues/1703


henri2h:
Works great!

PR #1741, opened 2026-04-01
feat: hotreload for freya core/winit

SergioRibera:
Adds support for hot-reloading via `subsecond` and `dioxus-devtools` behind a new `hotreload` feature flag.

When enabled, component functions are wrapped with `HotFn::current(...)` so Freya can pick up live code changes without restarting the app. A `subsecond` handler is registered on window creation to trigger a re-render on each patch.

Usage:
```
dx serve --hot-patch --features hotreload
```

Also updates the Nix dev shell with the dependencies required by `dioxus-cli`, and adds a `hr` recipe to the justfile for convenience.

[2026-04-01_17-11-23(1).webm](https://github.com/user-attachments/assets/b93bbb0c-48ff-4916-8b7d-1ae0c4638ed6)

### DIFF
--- a/crates/freya-core/src/element.rs
+++ b/crates/freya-core/src/element.rs
@@ -251,6 +251,7 @@ impl PartialEq for AppComponent {
     }
 }
 
+#[cfg(feature = "hotreload")]
 impl<F, E> From<F> for AppComponent
 where
     F: Fn() -> E + Clone + 'static,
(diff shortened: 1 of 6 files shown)


crates/freya-core/Cargo.toml
```diff
@@ -60,6 +60,8 @@ paste = "1.0.15"
 
 # Devtools
 serde = { version = "1", optional = true }
+subsecond = { version = "0.7.4", optional = true }
+dioxus-devtools = { version = "0.7.4", optional = true }
```
marc2332, line 64: Hold on... This dependency literally pulls dioxus-core and more, we seem to only be using it because of `connect_subsecond`... can we not rely on this dependency for that? https://docs.rs/dioxus-devtools/0.7.4/src/dioxus_devtools/lib.rs.html#76-86

marc2332: You will need the crate `dioxus_cli_config` for as long as we dont have our own CLI

SergioRibera: Oh yes, we are planning to have a separate crate that has all the types needed for hotreload, but that would require a custom CLI.

Wait, I think I can do it while preserving the websocket and keeping it working with `dioxus-cli`

PR #1751, opened 2026-04-05
Fixed show_progress for ProgressBar

zko-jin:
I'm assuming show_progress is supposed to hide the percentage label, from what I've seen the variable itself is not even being used anywhere in the code.


marc2332:
Seems like I overlooked this when I rewrote freya. Can you sign your commits btw?

marc2332 (approved):

PR #1763, opened 2026-04-07
fix(winit): Recreate swapchain if size doesnt match

marc2332:
This is kinda of failsafe in case the render events are not on sync (seems like it might be the case with windows when double clicking the title bar?)


marc2332:
has not proven to fix it. neither I could reproduce it tbh.

PR #1782, opened 2026-04-12
feat: Non-panicking version of queries invalidations

zko-jin:
Currently you can crash your application if the query being invalidated doesn't exist in the storage. This doesn't make much sense as some mutations can invalidate multiple queries some of which aren't being used currently and may not be initialized in the context.

I changed that behavior so we just ignore the request if the storage is not in context.

### DIFF
--- a/crates/freya-query/src/query.rs
+++ b/crates/freya-query/src/query.rs
@@ -332,7 +332,7 @@ impl<Q: QueryCapability> QueriesStorage<Q> {
     pub async fn invalidate_all() {
         let storage = consume_context::<QueriesStorage<Q>>();
 
-        storage.internal_invalidate_all().await;
+        storage.inner_invalidate_all().await;
     }
 
     /// Non-panicking version of [`QueriesStorage::invalidate_all()`]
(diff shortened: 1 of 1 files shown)


marc2332 (changes requested):
I am not sure if this is a good solution for a potentially non-problem.

In what case could you be invalidating a query where storage is not yet created? And would silently ignoring be a good solution?

I am not sure, perhaps yes.

zko-jin:
> I am not sure if this is a good solution for a potentially non-problem.
> 
> In what case could you be invalidating a query where storage is not yet created? And would silently ignoring be a good solution?
> 
> I am not sure, perhaps yes.

In my case I have queries that track and modify my torrents, one queries all torrents and the other queries specific torrents. 

There's a torrent client page that uses the ALL query and there are pages for specific content, let's say I have a page for manga X and I individually track each chapter. I can delete torrents in both the client page and on the manga page.

I can't merge both queries into one because one requires the info hash and for the other I don't know anything. What happens is that if I first enter the torrent client page and delete a torrent it tries to invalidate the individual query too and it ends up crashing.

I don't see a reason to just panic bcs the query was not found, perhaps we could add a try_invalidate function instead? If you give the ok I can make these changes

marc2332:
> I don't see a reason to just panic bcs the query was not found, perhaps we could add a try_invalidate function instead? If you give the ok I can make these changes

I like that better yep!

marc2332 (changes requested):
Replace `internal` with `inner`, rest lgtm

marc2332 (approved):

marc2332:
thanks!

PR #1787, opened 2026-04-13
fix(core): avoid accidental removal of old hash and SmallVec spill in LRUCache

LMQ0204:
This PR fixes two issues in the LRUCache `utilize` function:

1. **Accidental removal of old hash**: When a user's hash list had exactly 1 entry, calling `utilize` with a new second hash would incorrectly remove the first entry entirely, leaving only 1 hash. This caused the cached value to be evicted prematurely if it was only referenced by that user.

**Reproduction Test Case**:
```rust
#[test]
fn test_keep_old_hash_when_adding_second_hash() {
    let mut cache = LRUCache::<i32, u64>::default();

    cache
        .utilize(1, &50)
        .unwrap_or_else(|| cache.insert(1, &50, 5000));
    cache
        .utilize(2, &60)
        .unwrap_or_else(|| cache.insert(2, &60, 6000));
    assert_eq!(cache.utilize(1, &50), Some(Rc::new(5000)));
    assert_eq!(cache.utilize(2, &60), Some(Rc::new(6000)));

    cache
        .utilize(1, &60)
        .unwrap_or_else(|| cache.insert(1, &60, 6000));

    // ❌ Before fix: hash 50 was accidentally removed
    // ✅ After fix: both 50 and 60 are retained
    assert!(cache.get(&50).is_some());
    assert!(cache.get(&60).is_some());

    let hashes = cache.users.get(&1).unwrap();
    assert_eq!(hashes.len(), 2);
}
```
2. **Unnecessary SmallVec heap spill**: The original code would first push a 3rd hash (triggering a heap spill) and then drain back to 2. Since SmallVec does not automatically unspill, the data remained on the heap permanently.

**Verification Test Case**:
```rust
#[test]
fn test_smallvec_no_spill_on_third_hash() {
    let mut cache = LRUCache::<i32, u64>::default();

    cache
        .utilize(1, &50)
        .unwrap_or_else(|| cache.insert(1, &50, 5000));
    cache
        .utilize(1, &60)
        .unwrap_or_else(|| cache.insert(1, &60, 6000));

    let hashes = cache.users.get(&1).unwrap();
    assert!(!hashes.spilled());

    cache
        .utilize(2, &70)
        .unwrap_or_else(|| cache.insert(2, &70, 7000));
    cache
        .utilize(1, &70)
        .unwrap_or_else(|| cache.insert(1, &70, 7000));

    let hashes = cache.users.get(&1).unwrap();
    assert_eq!(hashes.len(), 2);
    // ✅ After fix: still not spilled, since we clean before pushing
    assert!(!hashes.spilled());
}
```

### Changes made:
- Adjusted the match branch from `len => len - 2` to `len => len - 1` to fix the logic bug
- Reordered operations to clean up old hashes *before* pushing new ones, ensuring the SmallVec length never exceeds 2
- Added two test cases:
  - `test_keep_old_hash_when_adding_second_hash`: Verifies the logic fix
  - `test_smallvec_no_spill_on_third_hash`: Verifies no heap spill occurs

### DIFF
--- a/crates/freya-core/src/lru_cache.rs
+++ b/crates/freya-core/src/lru_cache.rs
@@ -40,23 +40,16 @@ impl<V, ID: Hash + Eq> LRUCache<V, ID> {
         let mut hasher = FxHasher::default();
         hash_value.hash(&mut hasher);
         let hash = hasher.finish();
-        let mut value = self.map.get_mut(&hash);
 
-        let cache_value = value.as_ref().map(|v| v.1.clone());
+        let cache_value = self.map.get(&hash).map(|v| v.1.clone());
 
         let hashes = self.users.entry(id).or_default();
 
         // New hashed value
         if !hashes.contains(&hash) {
-            if let Some(value) = &mut value {
-                value.0 += 1;
-                hashes.push(hash);
-            }
-
             let index = match hashes.len() {
                 ..=1 => 0,
-                2 => 1,
-                len => len - 2,
+                len => len - 1,
             };
             // Clean the first current hash
             for old_hash in hashes.drain(0..index) {
@@ -69,6 +62,10 @@ impl<V, ID: Hash + Eq> LRUCache<V, ID> {
                 if entry.0 == 0 {
                     self.map.remove(&old_hash);
                 }
+            }
+            if let Some(value) = self.map.get_mut(&hash) {
+                value.0 += 1;
+                hashes.push(hash);
             }
         }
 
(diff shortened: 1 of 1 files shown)


marc2332 (changes requested):
Did you actually see how LRUCache is used?

`This caused the cached value to be evicted prematurely if it was only referenced by that user.`

This is not a bug, this is a feature. Values are only stored for as long as there is at least one user using it. If the user starts using another hash then the old one must be pruned. This is exclusively used for text caching where each element is using **1** paragraph at a time.

This is `N -> 1` relation.

marc2332:
Why did you try to fix this? Why were you looking specifically at this piece of code? Did you use AI?

LMQ0204:
> Why did you try to fix this? Why were you looking specifically at this piece of code? Did you use AI?

Sorry about the confusion earlier! My bad for completely misunderstanding the design intent.

To answer your questions:
I was just reading through the source code to learn how the text caching works, and ended up digging into this LRUCache part.
I tried to fix it because two things stood out to me: 

first, I saw the `SmallVec<[u64; 2]>` and the `len - 2` line, so I mistakenly thought it was supposed to keep the last 2 hashes per user. That's why I thought the 1->2 removal was a bug—totally my mistake.

 Second, I noticed the push-then-clean order causes a permanent SmallVec heap spill when you add a third hash, which seemed like an unnecessary performance hit even with the N->1 design.

And yeah, I only used AI to translate my thoughts into English because my English isn't great. All the code analysis, finding the issues, and writing the actual changes were done manually. 

I haven't updated the PR yet. I'll go fix it later to remove all the logic bug stuff, and only keep the fix for the SmallVec spill. Thanks for clarifying the design!

LMQ0204:
I'll close this PR for now since my current change implements N->2 behavior.

That said, I did want to note that even with the original code, there's still a SmallVec heap spill issue: when a user accesses a third hash, the SmallVec grows to 3 elements and spills to the heap permanently, since SmallVec doesn't automatically move back to inline storage after spilling.

I'm going to think about this a bit more first, to see if there's a better way to fix the spill issue without changing the intended N->1 behavior.

LMQ0204:
I was just re-reading the test case you have for `LRUCache`, and I wanted to ask a quick question to make sure I understand correctly.
```rust
#[test]
fn lru_cache() {
    let mut cache = LRUCache::<i32, u64>::default();

    cache
        .utilize(1, &50)
        .unwrap_or_else(|| cache.insert(1, &50, 5000));

    assert_eq!(cache.utilize(1, &50), Some(Rc::new(5000)));

    cache
        .utilize(1, &60)
        .unwrap_or_else(|| cache.insert(1, &60, 6000));
    assert_eq!(cache.utilize(1, &60), Some(Rc::new(6000)));
    assert_eq!(cache.utilize(1, &50), Some(Rc::new(5000)));

    cache
        .utilize(1, &70)
        .unwrap_or_else(|| cache.insert(1, &70, 7000));
    assert!(cache.get(&50).is_none());

    assert_eq!(cache.utilize(1, &60), Some(Rc::new(6000)));
    assert_eq!(cache.utilize(1, &70), Some(Rc::new(7000)));
    assert_eq!(cache.utilize(1, &60), Some(Rc::new(6000)));
    assert_eq!(cache.utilize(1, &70), Some(Rc::new(7000)));

    cache.remove(&1);
    assert!(cache.get(&60).is_none());
    assert!(cache.get(&70).is_none());

    cache
        .utilize(1, &70)
        .unwrap_or_else(|| cache.insert(1, &70, 7000));
    assert!(cache.utilize(2, &70).is_some());

    cache.remove(&1);
    assert!(cache.get(&70).is_some());
    cache.remove(&2);
    assert!(cache.get(&70).is_none());
}
```
If the intended design is truly N->1 (each user only keeps the single latest hash), then it seems like the test should fail at these points:

1.  After adding `60`, the test asserts that `50` is still accessible:
    ```rust
    assert_eq!(cache.utilize(1, &60), Some(Rc::new(6000)));
    assert_eq!(cache.utilize(1, &50), Some(Rc::new(5000))); // <-- This would be None if it's strictly N->1
    ```

2.  Later, when switching back and forth between `60` and `70`, the test expects both to remain accessible:
    ```rust
    assert_eq!(cache.utilize(1, &60), Some(Rc::new(6000)));
    assert_eq!(cache.utilize(1, &70), Some(Rc::new(7000)));
    assert_eq!(cache.utilize(1, &60), Some(Rc::new(6000))); // <-- This would be None if it's strictly N->1
    assert_eq!(cache.utilize(1, &70), Some(Rc::new(7000)));
    ```

This test behavior seems to perfectly match an N->2 design (keep the last 2 hashes), which is what my original PR implemented while also fixing the SmallVec heap spill.

Could you clarify if the test is actually demonstrating the intended behavior, or if there's something I'm missing here?

Thanks!

PR #1788, opened 2026-04-13
Add simple template and update flake

RobertasJ:
`nix flake init -t github:marc2332/freya` should now create a devshell env that supports freya codebases.

Made default devshell include no rust toolchain. 

Made stable devshell the same as default was before.


.gitignore
```diff
@@ -1,14 +1,14 @@
...
+/.idea
 snapshot_before.png
 snapshot_after.png
 documents_example
-bacon.toml
-.envrc
-.direnv
+/bacon.toml
+/.envrc
+/.direnv
```
SergioRibera, line 9: I think that of all the changes in gitignore, these are the ones that should be preserved; the rest don't make much sense 🤔

SergioRibera (changes requested):

.gitignore
RobertasJ: the bacon one is fine too i think, it's a config in workspace root

marc2332 (changes requested):
I am all in to update the flake but why exactly do we need this nix template for? I dont want to maintain that.

RobertasJ:
Its for being able to easily get a nix config for programming a freya app, just makes it so you dont have to go over the freya repo to copy it. 

If you dont want to maintain it then i will remove it

marc2332:
Closing in favor of simpler https://github.com/marc2332/freya/pull/1811

PR #1791, opened 2026-04-14
chore: Use `toolchain` files as single source of truth

sectore:
for Rust versions (default + nightly).

## Problem

Currently Rust versions (default + nightly) are defined all over the place. Check [PR #1638](https://github.com/marc2332/freya/pull/1638/changes) and you will know.

## Fix

Use two `toolchain` files for `default` and `nightly` as a single source of truth 

## Changes

- Add another `rust-toolchain-nightly.toml` (besides default `rust-toolchain.toml`)
- Extend `justfile` to get Rust versions from `toolchain` files.  
- In `.github/workflows/*.yml` files use [`actions-rust-lang/setup-rust-toolchain`](https://github.com/actions-rust-lang/setup-rust-toolchain) to handle different `toolchain` files (instead of hard-coded Rust versions)
- Ignore `.direnv` folder created by [`direnv`](https://direnv.net/) / Nix users
- flake.nix: Add missing dependencies, fix deprecated `xorg` packages
- Run `nix flake update` to get latest inputs, especially latest Rust versions.

## TODO

- [x] ~~Check if CI/GH workflows still work~~ DONE


marc2332 (changes requested):
Why did you make this specific change? Like, how did you end up finding this "issue", solving it and opening a PR? Because, if I have to be honest, this looks AI generated. I am in favor of AI, but against slop.

sectore:
As a Nix user I have to figure out how to run this project, especially by having different toolchain. This PR includes all changes I needed. It's not AI generated (no vibe-coding), but I have paired w/ Claude.

If you think it's slop to be more DRY, just ignore/close it. Changes are based on few hours work btw, AI would not do that in this way. Instead AI is still happy to put same versions/hashes everywhere and at different places (which is more than questionable). Call it human experiences vs. AI. Hope this PR makes more sense now.

flake.nix
```diff
@@ -11,20 +11,21 @@
...
     flake-utils.lib.eachDefaultSystem (
-      system:
-      let
+      system: let
         pkgs = nixpkgs.legacyPackages.${system};
 
         commonPackages = [
           pkgs.python3
+          pkgs.just
+          pkgs.taplo
```
marc2332, line 27: Duplicated
```suggestion
```

flake.nix
```diff
@@ -61,20 +64,34 @@
...
             buildInputs = commonBuildInputs;
             LD_LIBRARY_PATH = "${pkgs.lib.makeLibraryPath buildInputs}";
           };
-      in
-      {
+
+        stableToolchain = fenix.packages.${system}.fromToolchainFile {
+          file = ./rust-toolchain.toml;
+          # sha256 = pkgs.lib.fakeSha256;
+          sha256 = "sha256-zC8E38iDVJ1oPIzCqTk/Ujo9+9kx9dXq7wAwPMpkpg0=";
```
marc2332, line 83: can you leave a comment saying how to recalculate the hash or something?

.github/workflows/rust_sync.yml
```diff
@@ -20,8 +20,8 @@ jobs:
         with:
           fetch-depth: 2
           submodules: true
-      - name: Install stable rust toolchain
-        uses: dtolnay/rust-toolchain@aad518f59d88bae90133242f9ddac7f8bbc5dddf # 1.94
+      - name: Install Rust toolchain
+        uses: actions-rust-lang/setup-rust-toolchain@v1
```
marc2332: Please pin using a hash this and the rest of workflows.

.github/workflows/rust_lint.yml
```diff
@@ -87,5 +84,12 @@ jobs:
         save-if: ${{ github.ref == 'refs/heads/main' }}
     - name: Lint
       run: just c-ci
+    - name: Get nightly toolchain version
+      run: echo "NIGHTLY_TOOLCHAIN=$(just rv-nightly)" >> $GITHUB_ENV
+    - name: Install nightly Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: ${{ env.NIGHTLY_TOOLCHAIN }}
+        components: rustfmt
```
marc2332, line 95: For the sake of having the code pretty, can you move this nightly installation steps just below the stable one?     So like, have them together.

justfile
```diff
@@ -1,10 +1,22 @@
+toolchain := `sed -nr 's/channel = "(.*)"/\1/p' rust-toolchain.toml`
+nightly_toolchain := `sed -nr 's/channel = "(.*)"/\1/p' rust-toolchain-nightly.toml`
```
marc2332, line 2: Not that I use Windows but this effectively prevents from using windows, is there maybe another way? Otherwise I am fine with it

marc2332 (changes requested):

.github/workflows/rust_lint.yml
sectore: `nightly` needs to be installed just before `rustfmt` to force CI to use it. In other case CI will fail.

See commit message https://github.com/marc2332/freya/pull/1791/changes/dc727ad101fe2a7c201ea34341a102c7f6d19d6e _"install `nightly` toolchain after `lint` to avoid overriding `RUSTUP_TOOLCHAIN`"_

justfile
sectore: An option might be Just's built-in function [`os_family()`](https://just.systems/man/en/built-in-functions.html?highlight=os_family()#built-in-functions) to use another command for Windows users. Any suggestion for such command? _I'm not a Windows user...._

marc2332: Nah, it's fine, let's leave it like this

marc2332:
thanks!

PR #1823, opened 2026-04-28
fix(docs): update image generation timing

sectore:
## Description

After running `just t` few images have been re-generated. It turns out some of those are showing spinners, closed overlays or collapsed menus instead of expected content. Increase of poll duration in doc tests fixes it. 

Note: Few images have been updated because of small size changes. No issue here.
Note2: If you think a duration poll for generating `gallery_gif_viewer.png` is too high, using another `gif` with a smaller file size might help here.

## Checklist

- [x] I have tested my changes by running `just t`
- [x] I used AI assistance (Claude) to find relevant code


marc2332:
Yeah, I had this issue for a while but I procrastinated it, will check this asap

PR #1834, opened 2026-05-03
feat(components): Custom Cursor Icon for Button component

### LINKED ISSUE #1833: Allow changing cursor icon of builtin components (where CursorArea has no effect)
### Problem Description

<details><summary>First I wanted to say thank you for sharing this framework, I really like the approach :) Also that Freya takes accessibility into account </summary>
and that recent Release Candidate also allows to customize behaviour for many components by giving access to access-kit's methods (some components like `Input` don't expose an builder and  currently only allow setting `a11y_id`).
</details>

While testing Freya for one of my own projects, I noticed that ready to use components often define a fixed mouse icon and it's not possible to change them.

For example the Button component sets `Cursor::Pointer` which is not the usual behavour for native UI Buttons (expected: `Cursor::defaut()`) and much more common for Hyperlinks:

**button.rs:390-399**: (v0.4.0-rc.19)
```rust
  .on_pointer_enter(move |_| {
      if enabled() {
          Cursor::set(CursorIcon::Pointer);
      } else {
          Cursor::set(CursorIcon::NotAllowed);
      }
  })
  .on_pointer_leave(move |_| {
      Cursor::set(CursorIcon::default());
  })
```

Using the `CursorArea` component here causes overhead and does not allow changing the icon if the child component itself modifies the cursor. Probably because the next mouse enter immediatly overwrites the icon set by `CursorArea`.

### Proposed Solution

Allow changing mouse icons for builtin components with additional methods. Use `Cursor::default()` for buttons on mouse hovers, because this is the expected behaviour for non-web native UI buttons (Web-Browsers often treat `<button>` elements differently than native user interfaces.).

### Alternatives Considered

_No response_

### Additional Context

_No response_

Sebi2020:
## Description

This PR sets the usual mouse icon as a default for buttons and allows overriding the icon for components setting a cursor icon on mouse overs.

## Motivation
I noticed that buttons currently do not allow changing the icon and the existing `CursorArea` component does not allow changing the icon of a button. See #1833 

## Checklist

- [X] I have tested my changes.
- [X] I used AI assistance (e.g. ChatGPT, Copilot, etc.) to write this code.
- [X] I have added or updated documentation where relevant.

### DIFF
--- a/crates/freya-components/src/button.rs
+++ b/crates/freya-components/src/button.rs
@@ -140,6 +140,7 @@ pub struct Button {
     layout_variant: ButtonLayoutVariant,
     enabled: bool,
     focusable: bool,
+    cursor_icon: Option<CursorIcon>,
 }
 
 impl Default for Button {
@@ -173,6 +174,7 @@ impl Button {
             elements: Vec::default(),
             enabled: true,
             focusable: true,
+            cursor_icon: None,
             key: DiffKey::None,
         }
     }
@@ -260,6 +262,12 @@ impl Button {
     pub fn flat(self) -> Self {
         self.style_variant(ButtonStyleVariant::Flat)
     }
+
+    /// Override the cursor icon shown when hovering over the button while enabled.
+    pub fn icon(mut self, cursor_icon: impl Into<CursorIcon>) -> Self {
+        self.cursor_icon = Some(cursor_icon.into());
+        self
+    }
 }
 
 impl CornerRadiusExt for Button {
@@ -275,6 +283,7 @@ impl Component for Button {
         let focus_status = use_focus_status(focus);
 
         let enabled = use_reactive(&self.enabled);
+        let cursor_icon = self.cursor_icon.unwrap_or_default();
         use_drop(move || {
             if hovering() {
                 Cursor::set(CursorIcon::default());
@@ -389,7 +398,7 @@ impl Component for Button {
             })
             .on_pointer_enter(move |_| {
                 if enabled() {
-                    Cursor::set(CursorIcon::Pointer);
+                    Cursor::set(cursor_icon);
                 } else {
                     Cursor::set(CursorIcon::NotAllowed);
                 }
--- a/crates/freya-components/src/button.rs
+++ b/crates/freya-components/src/button.rs
@@ -140,7 +140,7 @@ pub struct Button {
     layout_variant: ButtonLayoutVariant,
     enabled: bool,
     focusable: bool,
-    cursor_icon: Option<CursorIcon>,
+    cursor_icon: CursorIcon,
 }
 
 impl Default for Button {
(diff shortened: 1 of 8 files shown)


crates/freya-components/src/button.rs
```diff
@@ -140,6 +140,7 @@ pub struct Button {
     layout_variant: ButtonLayoutVariant,
     enabled: bool,
     focusable: bool,
+    cursor_icon: Option<CursorIcon>,
```
marc2332:
```suggestion
    cursor_icon: CursorIcon,
```

crates/freya-components/src/button.rs
```diff
@@ -173,6 +174,7 @@ impl Button {
             elements: Vec::default(),
             enabled: true,
             focusable: true,
+            cursor_icon: None,
```
marc2332:
```suggestion
            cursor_icon: CursorIcon::default(),
```

crates/freya-components/src/button.rs
```diff
@@ -260,6 +262,12 @@ impl Button {
     pub fn flat(self) -> Self {
         self.style_variant(ButtonStyleVariant::Flat)
     }
+
+    /// Override the cursor icon shown when hovering over the button while enabled.
+    pub fn icon(mut self, cursor_icon: impl Into<CursorIcon>) -> Self {
+        self.cursor_icon = Some(cursor_icon.into());
```
marc2332:
```suggestion
    pub fn cursor_icon(mut self, cursor_icon: impl Into<CursorIcon>) -> Self {
        self.cursor_icon = cursor_icon.into();
```

crates/freya-components/src/button.rs
```diff
@@ -275,6 +283,7 @@ impl Component for Button {
         let focus_status = use_focus_status(focus);
 
         let enabled = use_reactive(&self.enabled);
+        let cursor_icon = self.cursor_icon.unwrap_or_default();
```
marc2332:
```suggestion
        let cursor_icon = self.cursor_icon;
```

marc2332 (changes requested):
some improvements, thanks for the pr!

marc2332:
Do you feel like doing this same change for the rest of the components that also change the cursor? Feel free to decline, it is just a suggestion. If you decide to do it you can do all of them in this same PR, no need to create more :)

Sebi2020:
> Do you feel like doing this same change for the rest of the components that also change the cursor? Feel free to decline, it is just a suggestion. If you decide to do it you can do all of them in this same PR, no need to create more :)

I tried to cover all components which set an icon inside the components crate. I was not sure if the defaults should be kept for the other components or if they should also use the default cursor.

marc2332:
> I was not sure if the defaults should be kept for the other components or if they should also use the default cursor.

Let's switch them all to `CursorIcon::default()` too :)

marc2332 (approved):

PR #1838, opened 2026-05-03
feat(menu): close on `Escape`

sectore:
## Description

- Close menu/submenu on `ESC`. One at time.
-  Fix `examples/feature_material_design` example: menu requires two clicks/escapes to close (instead of one)

## Motivation

Similar to other desktops apps, e.g. Firefox.

## Checklist

- [x] I used AI assistance (CLAUDE)


marc2332 (approved):
Nice catch

PR #1884, opened 2026-05-12
feat: impl IntoReadable for Memo

Zomatree:
## Description

Adds support for converting `Memo<T>` into a `Readable<T>`, useful for passing down the memoized value to other views.

## Motivation

Currently this must be done manually via the `Readable::new` constructor which adds unnecessary boilerplate.

## Checklist

- [x] I have added or updated documentation where relevant.

### DIFF
--- a/crates/freya-core/src/lifecycle/memo.rs
+++ b/crates/freya-core/src/lifecycle/memo.rs
@@ -59,6 +59,12 @@ impl<T: Copy + PartialEq + 'static> Deref for Memo<T> {
     }
 }
 
+impl<T> Memo<T> {
+    pub fn as_state(&self) -> State<T> {
+        self.state
+    }
+}
+
 impl<T: PartialEq> Memo<T> {
     /// Adapted from https://github.com/DioxusLabs/dioxus/blob/a4aef33369894cd6872283d6d7d265303ae63913/packages/signals/src/read.rs#L246
     /// SAFETY: You must call this function directly with `self` as the argument.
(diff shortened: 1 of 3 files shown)


crates/freya-core/src/lifecycle/memo.rs
```diff
@@ -59,6 +59,12 @@ impl<T: Copy + PartialEq + 'static> Deref for Memo<T> {
     }
 }
 
+impl<T> Memo<T> {
+    pub fn as_state(&self) -> State<T> {
+        self.state
+    }
+}
```
marc2332:
```suggestion
```

crates/freya-core/src/lifecycle/readable.rs
```diff
@@ -135,8 +135,20 @@ impl<T: 'static> IntoReadable<T> for T {
     }
 }
 
+impl<T: 'static> IntoReadable<T> for Memo<T> {
+    fn into_readable(self) -> Readable<T> {
+        Readable::from_state(self.as_state())
```
marc2332:
```suggestion
        Readable::from_state(self.state)
```

marc2332 (changes requested):

crates/freya-core/src/lifecycle/memo.rs
marc2332: I dont want such method public :)

marc2332 (approved):

PR #1905, opened 2026-05-22
Anonymous contribution via gitGost

gitgost-anonymous:
warnings in tests

---

*This is an anonymous contribution made via [gitGost](https://gitgost.leapcell.app).*

*The original author's identity has been anonymized to protect their privacy.*


fengints:
this line actually have meaning.
At least example build won't fail if user uses simple `git clone`

PR #1913, opened 2026-05-24
feat(components): Smart positioning for `Select`

### LINKED ISSUE #1912: enhancement: dynamic dropdown direction
### Problem Description

Right now it seems that dropdowns only drop... down. This can be a problem when there are a lot of elements, and it might make more sense to flow upwards instead of down, like what browsers do with the right-click menu. Here is what it looks like for my case right now:

Correct:
Image

Incorrect:
Image

### Proposed Solution

Automatic overflow detection or some sort of `if element_list.y > screen_height/2 { ... }` check.

### Alternatives Considered

_No response_

### Additional Context

_No response_

marc2332:
Closes https://github.com/marc2332/freya/issues/1912

### DIFF
--- a/crates/freya-components/src/menu.rs
+++ b/crates/freya-components/src/menu.rs
@@ -6,6 +6,7 @@ use torin::{
         Alignment,
         Area,
         Position,
+        Size2D,
     },
     size::Size,
 };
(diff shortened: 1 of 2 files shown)


SpikeHD:
Tested and working :rocket:

PR #1982, opened 2026-06-29
chore(components): Smart positioning for ColorPicker

### LINKED ISSUE #1981: Color picker draws outside the screen
### Problem Description

Much like the previous issue I've opened for dropdowns (#1912), the ColorPicker component always opens to the right, off the screen. It also doesn't open absolutely positioned as far as I can tell, but relatively positioned, and it causes a layout shift as a consequence to that

Image

### Proposed Solution

The color picker should be positioned absolutely, and positioned in a way that it cannot overflow out of the window.

### Alternatives Considered

_No response_

### Additional Context

_No response_

marc2332:
closes https://github.com/marc2332/freya/issues/1981

### DIFF
--- a/crates/freya-components/src/color_picker.rs
+++ b/crates/freya-components/src/color_picker.rs
@@ -28,7 +28,6 @@ use crate::{
     menu::{
         Menu,
         MenuButton,
-        overflow_offset,
     },
 };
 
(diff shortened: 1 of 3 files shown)


marc2332:
CI failing because of https://github.com/rwf2/cookie-rs/pull/256

PR #2005, opened 2026-07-07
fix(components): Synchronous decoding by default in SVG, allow passing event handlers and use visible area for sizing, exclude margin from image rendering, prevent a11y init races

marc2332:
Closes https://github.com/marc2332/freya/issues/2004

- Synchronous decoding by default in SvgViewer
- Allow passing event handlers in SvgViewer
- Use visible area for sizing in SvgViewer
- Exclude margin from image element rendering
- Prevent a11y init races by guarding properly

PR #2006, opened 2026-07-07
fix(core):Implement event handling for ImageElement

LMQ0204:
## Description

`ImageElement` was missing an override of `events_handlers()` from `ElementExt`. The default implementation returns `None`, so the tree never registered the Image node's event handlers in its listener map. Consequently, any event handler attached to an `image()` element — including those forwarded by `SvgViewer` via `.with_event_handlers()` — was silently ignored and never fired.

This commit adds the missing `events_handlers()` override and also adds the `EVENT_HANDLERS` flag check in `diff()` so that handler changes on an existing Image element are properly tracked.

## Motivation

I encountered this bug while running the `component_input_form` example: clicking the eye icon had no effect — the password visibility toggle never fired. After investigation, the root cause was traced to `ImageElement` not reporting its event handlers to the tree.

## Checklist

- [ ] I have added or updated documentation where relevant.

### DIFF
--- a/crates/freya-core/src/elements/image.rs
+++ b/crates/freya-core/src/elements/image.rs
@@ -240,6 +240,10 @@ impl ElementExt for ImageElement {
             diff.insert(DiffModifies::STYLE);
         }
 
+        if self.event_handlers != image.event_handlers {
+            diff.insert(DiffModifies::EVENT_HANDLERS);
+        }
+
         diff
     }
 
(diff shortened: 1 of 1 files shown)


marc2332:
Done already in https://github.com/marc2332/freya/pull/2005/changes/47de05c76ed5bab1fefee72c646a348234120981

marc2332:
Such a coincidence, I literally pushed the change 1min ago, are you tracking my commits?

LMQ0204:
> Such a coincidence, I literally pushed the change 1min ago, are you tracking my commits?

I missed your commit — should have refreshed first. Appreciate you closing the PR.

PR #2081, opened 2026-07-29
feat(freya): Opt-in software-only renderer

marc2332:
Allow shipping apps with only a software renderer.


inthehack:
It's working for me. Thx very much 👍🏼 !

PR #2116, opened 2026-08-01
Swap alacritty_terminal for rio-vt in freya-terminal

raphamorim:
This swaps freya-terminal's emulation backend from alacritty_terminal to [rio-vt](https://crates.io/crates/rio-vt), the terminal core we extracted from [Rio](https://github.com/raphamorim/rio). portable-pty stays as the PTY layer, so the change is contained to the freya-terminal crate.

What changed:

- `pty.rs`: `Term` becomes `rio_vt::crosswords::Crosswords`; the `TermSize` shim is gone since rio-vt's size type implements its dimensions trait directly. `EventProxy` maps the same four events (`PtyWrite`, `Title`, `ResetTitle`, `ClipboardStore`).
- New `cell.rs`: rio-vt packs cell styling into per-grid side tables rather than inline on each cell, so visible rows are resolved into a small crate-local `TermCell` (char, colors, inverse/wide flags, zero-width chars, hyperlink URI) once per frame — mirroring the row cloning the renderer already did. `rendering.rs`/`url.rs` consume that instead of alacritty's `Cell`.
- `handle.rs`/`parser.rs`: `TermMode` → `Mode`, `Point` → `Pos`; selection, scrolling, and the mouse encoders keep identical logic (all the mode bits incl. the `MOUSE_MODE` composite exist under the same names).
- `colors.rs`: same palette logic; named-color variants are spelled `Light*` instead of `Bright*`.
- `osc7.rs`: unchanged behavior and tests, now driving rio-vt's parser (`rio_vt::performer::parser::{Parser, Perform}`). Worth noting rio-vt also tracks OSC 7 natively (`Crosswords::current_directory`), so the side channel could be dropped entirely in a follow-up if you prefer.
- `SelectionType` in the prelude now re-exports from rio-vt — same variants, so `feature_terminal`/`feature_terminal_canvas` compile unchanged.

Why: rio-vt is actively maintained, safe Rust, same alacritty lineage for grid/escape semantics, correct unicode width + grapheme clustering, and it slims the tree (alacritty_terminal's default features pull serde + winit-adjacent deps the widget never uses). It also opens doors for the terminal component down the line: the core decodes sixel/kitty/iTerm2 inline images behind a feature flag, has regex search over screen + scrollback, and parses OSC 133 semantic prompts. Benchmarks against alacritty_terminal and vt100: https://github.com/raphamorim/rio-vt-benchmark.

Verified: the crate's tests pass (OSC 7 suite + doctests), clippy is clean, and both terminal examples build. I couldn't do a visual pass on the examples here, so a quick run of `feature_terminal` on your side would be a good sanity check — happy to iterate on anything that looks off.


marc2332:
Nice work @raphamorim, I might merge it but I noticed this issue as well https://github.com/raphamorim/rio/issues/1753

marc2332:
I think the PR linked to that issue is not enough as we are using the rio-vt crate here.

marc2332:
Created a PR https://github.com/raphamorim/rio/pull/1788

raphamorim:
Noice will look in a bit 🙏

raphamorim:
@marc2332 published rio-vt 0.5.5 , lemme know if you want me to update here. thanks for the pr as well!!

raphamorim:
👀

marc2332:
Thanks @raphamorim ! 🦀

raphamorim:
Really cool!

PR #2125, opened 2026-08-02
fix(core): Render cursor when text is completely empty

marc2332:
Literally what the title says


marc2332:
Ended up also fixing some aligning issues of the cursor when there was no text and deduplicating logic

PR #2133, opened 2026-08-05
feat(winit): set main window title as a11y label

bitfehler:
## Description

Currently, main windows do not have any name in the accessibility tree. This commit makes -winit set a new Tree property to the title of the window, which is then used when constructing the a11y tree.

Note that this will not pick up changes to the window title performed after the window has been created.

## Motivation

I've been playing with accessibility in Freya, and given the general state of accessibility at least on Linux it's already pretty promising, thank you for that!

One thing I immediately noticed was that the main window does not have an accessibility label (that is, the screen reader doesn't know the name/title of the window). I initially thought that this should probably be handled by `accesskit_winit`, but indeed it isn't, as is also manifested in its ["simple" example](https://github.com/AccessKit/accesskit/blob/main/adapters/winit/examples/simple.rs#L115).

So I set out to add it in Freya, but honestly I might need a pointer or two. The window node gets its a11y attributes [here](https://github.com/marc2332/freya/blob/main/crates/freya-core/src/accessibility/tree.rs#L364), but I think getting the window title at this point is a no-go, as this would make -core depend on -winit.

What I did here was sort of the least-invasive change I could come up with. It's probably not great, but the question that looms over all of this is: if it is to be done right, should there be a "freya" way to change the window title? Right now this would go through `Platform::get().with_window()` which operates on a `winit` window, so this cannot be intercepted by Freya. I guess it would also be fine to say if you change the window title you have to change the a11y label yourself as well, but as is this isn't really easy either. So, any thoughts you might have around this would be greatly appreciated...

Edit: forgot to mention, but I suppose if there were to be a "freya" way to set/get the window title, it would probably be via `Platform` as well?

## Checklist

- [ ] I have added or updated documentation where relevant.

### DIFF
--- a/crates/freya-core/src/accessibility/tree.rs
+++ b/crates/freya-core/src/accessibility/tree.rs
@@ -1,4 +1,7 @@
-use std::any::Any;
+use std::{
+    any::Any,
+    borrow::Cow,
+};
 
 use accesskit::{
     Action,
@@ -362,6 +365,7 @@ impl AccessibilityTree {
 
         if node_id == NodeId::ROOT {
             accessibility_data.builder.set_role(Role::Window);
+            accessibility_data.builder.set_label(Cow::from(&tree.title));
         }
 
         // Set children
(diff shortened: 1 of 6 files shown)


marc2332:
Thanks for the PR @bitfehler :) I pushed some changes that I think improve it, lmk your thoughts :)

bitfehler:
I love it! :heart: :100: Thanks!

PR #2210, opened 2026-08-29
perf(core): Skip noop opacity layers and join the rest

### LINKED ISSUE #925: optimization: collect composition effects before applying them
Certain properties such as `opacity` and `transform` are applied to an element based on the state of its parents. For example, a parent with `opacity: 0.5` must propagate its opacity down to children.

Normally this effect is achieved through creating a new canvas layer with a certain opacity when a parent node has opacity < 1.0, then merging that layer back into the canvas after rendering the children, but this isn't possible with Freya's compositor due to the layering system resulting children of a parent node being potentially rendered after other siblings of the parent are drawn (freya draws layers from the bottom up rather than recursively drawing parent -> child).

To get around this, we currently propagate a list of opacity/rotation values down the node tree in `TransformState` as a `Vec<f32>`. This is then looped over in the rendering pipeline and applied to each child manually.

https://github.com/marc2332/freya/blob/dab31780609f76533314296ca8987c0b5ad0af3a/crates/core/src/render/pipeline.rs#L220-L231

The potential problem here is that a new canvas layer is created for each parent opacity value, which has some potential overhead (i'm not sure exactly how much, but each layer save requires skia to perform compositing). Rather than saving a new layer for each parent node's opacity, we can instead do one of the following:
- **Option A (less efficient, but easier):** Calculate the final opacity of the child node ahead of time, and only save a single layer per child with that opacity once.
- **Option B (more efficient, requires possible refactor):** Ensure that children of a parent node are rendered immediately after the parent if they are on the same layer. Call `save_layer_alpha_f()` when the parent is drawn, then `restore()` after drawing the last child, which will prevent the need to allocate a `Vec<f32>` for each node and also cut down on layer saves by only requiring one for the parent rather than one for each child. 
	- This would require a refactor to how layers are assigned. Children would need to be on the same layer as the parent unless otherwise specified and rely on `NodeId` ordering to be drawn later.

marc2332:
Closes #925

### DIFF
--- a/crates/freya-core/src/render_pipeline.rs
+++ b/crates/freya-core/src/render_pipeline.rs
@@ -167,8 +167,12 @@ impl RenderPipeline<'_> {
                         }
                     }
 
-                    for opacity in effect_state.opacities.iter() {
-                        self.canvas.save_layer_alpha_f(layer_bounds, *opacity);
+                    let opacity = effect_state
+                        .opacities
+                        .iter()
+                        .fold(1., |acc, opacity| acc * opacity);
+                    if opacity < 1. {
+                        self.canvas.save_layer_alpha_f(layer_bounds, opacity);
                     }
 
                     // Transform the canvas area given the scale effects


tropicaaal:
believe this addresses part of #925

marc2332:
> believe this addresses part of #925

True!

PR #2266, opened 2026-09-11
chore(components): Make canvas element clip by default

marc2332:
better dx!


inthehack (commented):
Looks good to me. Thx 🙏🏼

